PR #33, opened 2018-08-20
Use Afero Filesystem

pjdufour-truss:
This Pull Request adds support for the [Afero Filesystem](https://github.com/spf13/afero), which provides the capability to build and merge multiple PDF files in-memory.  This capability is incredibly useful for some web server scenarios.

The default is to still use the operating system file system, which should lead to zero changes in the CLI.  This PR passed the tests in the sub-packages `pdfcpu` and `api`.  Afero works on Mac OSX, Linux, and Windows, but there could be issues with other operating systems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 # pdfcpu: a golang pdf processor
 
 [![Build Status](https://travis-ci.org/hhrutter/pdfcpu.svg?branch=master)](https://travis-ci.org/hhrutter/pdfcpu)
-[![GoDoc](https://godoc.org/github.com/hhrutter/pdfcpu?status.svg)](https://godoc.org/github.com/hhrutter/pdfcpu)
+[![GoDoc](https://godoc.org/github.com/trussworks/pdfcpu?status.svg)](https://godoc.org/github.com/trussworks/pdfcpu)
 [![Coverage Status](https://coveralls.io/repos/github/hhrutter/pdfcpu/badge.svg?branch=master)](https://coveralls.io/github/hhrutter/pdfcpu?branch=master)
-[![Go Report Card](https://goreportcard.com/badge/github.com/hhrutter/pdfcpu)](https://goreportcard.com/report/github.com/hhrutter/pdfcpu)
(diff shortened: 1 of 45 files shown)


hhrutter:
Thanks for the PR!
Just out of curiosity, which use case are you solving with this?
I am not gonna lie that I am hesitant adding new dependencies other than package/errors,
I want to keep pdfcpu as close as can be to the stdlib.
Also I am working on a new release and need to see how this fits into the overall picture.
Let me get back to you.

pjdufour-truss:
Our primary use case is to dynamically generate a single downloadable PDF file that's a merge of images from AWS S3.  This needs to happen in a managed container environment without storing intermediate files (which could include PII) on disk.

hhrutter:
Sorry for the stupid questions :(

Let me see if I understand this and please correct me if I am wrong:

The functionality you are really looking for is in-memory pdf merging.
You need to be able to produce the resulting pdf in-memory.

Is it also a requirement to read the input pdf's from memory or is that optional for you?
What are these `images` you are referring to, are these pdf's pulled down from AWS S3? 

I understand you want to avoid leaving PII on disk in the process but what do you mean by `intermediary files`? When you are merging pdfs with pdfcpu there are no intermediary files.

pjdufour-truss:
By images we mean PDFs or JPEGs from AWS S3.  We download them and immediately add them to an in-memory filesystem.  They never touch disk.  It's those intermediate files that we'd be worried about cleaning up and I was not suggesting pdfcpu itself would leave anything behind.  However, the ability to read from an afero in-memory file system is what makes the integration seamless.

hhrutter:
Interesting use case and I am glad pdfcpu serves you well.
Just curious.. Anything more you can share about your usage of pdfcpu?
Maybe the industry you are in and are you using pdfcpu in production?

Ideally I'd like to get rid of usage of files across the board and instead use Reader/Writer.
That way pdfcpu would offer maximum flexibility digesting files, network streams and would also be working within an afero layer. Only getting that done ain't easy. 😞 

I get where you are coming from and understand your solution but I do not want to add any more non stdlib dependencies. pdfcpu is a lightweight package and I'd like to keep it that way.

Thank you for using pdfcpu.

hhrutter:
Your fork is an excellent tool for anybody using Afero.
Still, pdfcpu aims for minimum dependencies, so I have to close this.

PR #43, opened 2018-09-21
updated README.md

antzshrek:


hhrutter:
I would like to keep the README like it is.
Thanks so much for your input!

PR #46, opened 2018-10-04
Simplify removeAttachedFiles function

SimePel:


pkg/pdfcpu/attach.go
```diff
@@ -197,57 +197,55 @@ func removeAttachedFiles(xRefTable *XRefTable, files StringSet) (ok bool, err er
...
+	if len(files) == 0 {
+		err = xRefTable.RemoveEmbeddedFilesNameTree()
+		if err != nil {
+			return false, err
+		}
+		return true, nil
+	}
 
-		for fileName := range files {
+	removed := false
```
hhrutter: var removed bool .. initializes to false
I would like to keep this here.

SimePel: Ok, as you wish!

SimePel:
I understood. I will open this PR later.

hhrutter:
Thanks 💚

PR #47, opened 2018-10-29
ExtractImageData not always return nil when not filter.CCITTFax

ungerik:
The (working) filter.Flate case further down would never be reached without the change.

Also: Should an error be returned instead of nil, nil when filters are not supported?


hhrutter:
Oops, you're right!
Thanks for catching this late night bug 💚 

As in regards to what to do when a filter is not supported during image extraction I'll have to think about it. Actually right now this should never happen because there is a case for each filter supporting images I believe. The DCT and JPX cases need actual decoding but at least they return smth which is useful in most cases.

Issue #98, opened 2019-08-02
Error merging files during optimization


plaisted:
Encountered an error testing the tool for merging some PDFs I have.  The PDF did validate using your tool (and others).

```
 INFO: 2019/08/01 22:55:57 optimizing fonts & images
Fatal: equalObjects: unhandled compare for type <nil>

github.com/hhrutter/pdfcpu/pkg/pdfcpu.equalObjects
        /Users/horstrutter/go/src/github.com/hhrutter/pdfcpu/pkg/pdfcpu/equal.go:67
github.com/hhrutter/pdfcpu/pkg/pdfcpu.equalArrays
        /Users/horstrutter/go/src/github.com/hhrutter/pdfcpu/pkg/pdfcpu/equal.go:81
github.com/hhrutter/pdfcpu/pkg/pdfcpu.equalObjects
        /Users/horstrutter/go/src/github.com/hhrutter/pdfcpu/pkg/pdfcpu/equal.go:64
github.com/hhrutter/pdfcpu/pkg/pdfcpu.equalDicts
        /Users/horstrutter/go/src/github.com/hhrutter/pdfcpu/pkg/pdfcpu/equal.go:181
github.com/hhrutter/pdfcpu/pkg/pdfcpu.equalStreamDicts
        /Users/horstrutter/go/src/github.com/hhrutter/pdfcpu/pkg/pdfcpu/equal.go:96
github.com/hhrutter/pdfcpu/pkg/pdfcpu.handleDuplicateImageObject
        /Users/horstrutter/go/src/github.com/hhrutter/pdfcpu/pkg/pdfcpu/optimize.go:335
github.com/hhrutter/pdfcpu/pkg/pdfcpu.optimizeXObjectResourcesDict
        /Users/horstrutter/go/src/github.com/hhrutter/pdfcpu/pkg/pdfcpu/optimize.go:408
github.com/hhrutter/pdfcpu/pkg/pdfcpu.optimizeResources
        /Users/horstrutter/go/src/github.com/hhrutter/pdfcpu/pkg/pdfcpu/optimize.go:493
github.com/hhrutter/pdfcpu/pkg/pdfcpu.parseResourcesDict
        /Users/horstrutter/go/src/github.com/hhrutter/pdfcpu/pkg/pdfcpu/optimize.go:520
github.com/hhrutter/pdfcpu/pkg/pdfcpu.parsePagesDict
[...]

PR #106, opened 2019-08-29
Make LastModified optional rather than required in validatePieceDict

richardwilkes:
Many PDFs I've run across lately fail to load due to this being required
and changing it to optional allows them to load fine. LastModified does
not seem to be essential and allowing it to be optional should not be
harmful in any way.

### DIFF
--- a/pkg/pdfcpu/validate/xReftable.go
+++ b/pkg/pdfcpu/validate/xReftable.go
@@ -526,7 +526,11 @@ func validatePieceDict(xRefTable *pdf.XRefTable, d pdf.Dict) error {
 			continue
 		}
 
-		_, err = validateDateEntry(xRefTable, d1, dictName, "LastModified", OPTIONAL, pdf.V10)
+		required := REQUIRED
+		if xRefTable.ValidationMode == pdf.ValidationRelaxed {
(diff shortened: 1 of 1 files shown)


hhrutter:
Hey! Interesting!
I haven't seen that with my testsuite. 
Is there a way you can share a sample?
Regardless I would follow pdfcpu's tradition here and extend relaxed validation.
eg. like in validate/font.go on lines 378, 388

richardwilkes:
Sounds good. I've updated the change as suggested.

Unfortunately, I can't share any of the PDFs I've run across that have had this issue, as they are commercial products. Essentially, they are battle maps I bought for a role-playing game I play, which I'm extracting the images from so that I can manipulate them in a virtual table top (rather than printing and cutting them out, like I used to do).

Turns out I've been running into other issues with some of these as well, such as multiple filters, where pdfcpu only supports one. If I find the time and am able to fix it, I'll make another PR that addresses that, too.

hhrutter:
Excellent!
I've been wanting to get a hand on PDFs with not trivial filter pipelines!
This ties into #98 
Thank you so much for your contribution! 💚

PR #121, opened 2019-10-24
constant export

minenok-tutu:
When using pdfcpu via API, you need those constants to set watermark settings

### DIFF
--- a/pkg/pdfcpu/stamp.go
+++ b/pkg/pdfcpu/stamp.go
@@ -42,16 +42,16 @@ const (
 
 // Rotation along one of 2 diagonals
 const (
-	noDiagonal = iota
-	diagonalLLToUR
-	diagonalULToLR
+	NoDiagonal = iota
+	DiagonalLLToUR
+	DiagonalULToLR
 )
 
 // Render mode
 const (
-	rmFill = iota
-	rmStroke
-	rmFillAndStroke
+	RMFill = iota
+	RMStroke
+	RMFillAndStroke
 )
 
 var (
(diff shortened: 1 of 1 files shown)


hhrutter:
Hello!

Can you please give me an example of your use case?

The following does not work for you?

```
wmConf := "Test, d:1"
wm, _ := pdf.ParseWatermarkDetails(wmConf, true)
if err := AddWatermarksFile(inFile, outFile, selectedPages, wm, nil); err != nil {
   ...
}
```

minenok-tutu:
> Hello!
> 
> Can you please give me an example of your use case?
> 
> The following does not work for you?
> 
> ```
> wmConf := "Test, d:1"
> wm, _ := pdf.ParseWatermarkDetails(wmConf, true)
> if err := AddWatermarksFile(inFile, outFile, selectedPages, wm, nil); err != nil {
>    ...
> }
> ```

Oh, that was not obvious. I think, this is much more convenient:
```
wm := pdfcpu.DefaultWatermarkConfig()
wm.OnTop = true
wm.Scale = 0.05
wm.Dx = 200
wm.Dy = -18
wm.Diagonal = pdfcpu.NoDiagonal
pdfcpuapi.AddWatermarksFile(file, file, nil, wm, pdfcpu.NewDefaultConfiguration())
```
Because I need a structure for watermark config, so I look for some straight init. Doing it your way, I will need to learn how to fill this string correctly. And it is difficult to find.

hhrutter:
👍 Please sign the [Developer Certificate of Origin](https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin) so I can merge in this PR.

minenok-tutu:
Done :)

PR #123, opened 2019-10-30
Introduce function PageCountFile

matbur:
Hi!
First of all, I really like the idea of the project you are building here, and thank you for doing that!

We, at [Ingrid](ingrid.com), use this package to merge PDFs and we need to check if the number of pages in tests.

This PR introduces function `PageCountFile` which counts pages in `inFile`. Also function `PageCount` now expects `io.ReadSeeker`.

It requires to call `validate.XRefTable` inside `api.PageCount`. I tried to put this into `api.ReadContext` but then test `cli.TestOptimize` failed with case when `ValidationMode` was set to `pdf.ValidationNone` and I'm not able to fix this. Maybe you can help.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -64,7 +64,7 @@ func logOperationStats(ctx *pdf.Context, op string, durRead, durVal, durOpt, dur
 	ctx.Write.LogStats()
 }
 
-// ReadContext uses an io.Readseeker to build an internal structure holding its cross reference table aka the Context.
+// ReadContext uses an io.ReadSeeker to build an internal structure holding its cross reference table aka the Context.
 func ReadContext(rs io.ReadSeeker, conf *pdf.Configuration) (*pdf.Context, error) {
 	return pdf.Read(rs, conf)
 }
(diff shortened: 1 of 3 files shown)


hhrutter:
Hello there!

Thanks for putting in the effort.
Can you please state your exact problem?

Why do we need both
 `func PageCount(rs io.ReadSeeker, conf *pdf.Configuration) (int, error)`
and
`func PageCountFile(inFile string) (int, error) `?

Can't you just call `ctx.PageCount()` ?

hhrutter:
Without fully understanding your usecase I think it makes a lot of sense
to provide both ReadSeeker and filename based functions for getting the page count.
This is also serves consistency across the api.

What do you think about the following?
If it's ok with you, just patch your PR accordingly and I am happy to merge:

```
// PageCount returns rs's page count.
func PageCount(rs io.ReadSeeker, conf *pdf.Configuration) (int, error) {
	ctx, err := ReadContext(rs, conf)
	if err != nil {
		return 0, err
	}
	if err := ValidateContext(ctx); err != nil {
		return 0, err
	}
	return ctx.PageCount, nil
}

// PageCountFile returns inFile's page count.
func PageCountFile(inFile string) (int, error) {
	f, err := os.Open(inFile)
	if err != nil {
		return 0, err
	}
	defer f.Close()

	return PageCount(f, pdf.NewDefaultConfiguration())
}
```

matbur:
@hhrutter thanks for your response. Your snippet looks much better, so I committed it.

In function `api.ReadContextFile` context is validated but in `api.ReadContext` is not. I would like to move validation to `api.ReadContex` and remove it from `api.PageCount`. But then test `cli.TestOptimize` fails. What do you think about it?

hhrutter:
OK!
Couple of suggestions:

1) Please revert the change in cli_test.go line 138.
Over there is no bug, the intention is to show the variations of arguments to inFile & outFile. This is also unrelated to this issue.

2) Please adapt TestPageCount in api.go so it tests exactly 1 file of the testsuite and compares to the known page count. It does not make much sense to compare against something that gets called anyway.

3) Please sign your commit.

hhrutter:
OK!

Please squash your changes into a signed commit so I can merge this in.

https://help.github.com/en/github/authenticating-to-github/signing-commits

matbur:
@hhrutter I think I've done everything. Thanks for your help with this.

PR #127, opened 2019-11-10
Fix Color Lookup for Flat encoded image

dharnitski:
Fix Images extraction bug for documents with Flat Encoded Image and Indexed CS.

1. For Lookup Colors Unescape String Literal. Currently we only convert sting to bytes.
2. Unescape String - fix Octal String encoding.


hhrutter:
Signing commits: https://help.github.com/en/github/authenticating-to-github/signing-commits

PR #162, opened 2020-01-23
Fixed a bug in paperSize.go toMillimetres, which was returning d.Heig…

cewitte:
…ht twice instead of d.Width and d.Height; and fixed ToMillimeters and ToCentimeters spelling errors.

### DIFF
--- a/pkg/pdfcpu/paperSize.go
+++ b/pkg/pdfcpu/paperSize.go
@@ -56,7 +56,7 @@ func (d Dim) ToCentimetres() Dim {
 
 // ToMillimetres converts d to centimetres.
 func (d Dim) ToMillimetres() Dim {
-	return Dim{d.Height * userSpaceToMm, d.Height * userSpaceToMm}
+	return Dim{d.Width * userSpaceToMm, d.Height * userSpaceToMm}
 }
 
 // AspectRatio returns the relation between width and height.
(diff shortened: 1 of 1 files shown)


pkg/pdfcpu/paperSize.go
```diff
@@ -49,14 +49,14 @@ func (d Dim) ToInches() Dim {
 	return Dim{d.Width * userSpaceToInch, d.Height * userSpaceToInch}
 }
 
-// ToCentimetres converts d to centimetres.
-func (d Dim) ToCentimetres() Dim {
+// ToCentimeters converts d to centimetres.
```
hhrutter: I am ok with using the American spelling across the board for centimeters and millimeters, but this has to be done consistently for all occurences in the code base including all comments.

For function name changes please also fix all references eg. like in context.go

cewitte: Thanks for your comments.

I was looking into it and there are a lot of cases like that, from switch/case scenarios to func calls in many places. I could make the changes, but there's a huge risk of breaking too much of other people's code. From godoc.org it felt like an isolated thing, but in fact it's not (sorry for the lack of proper investigation).

Please feel free to reject the changes - the small bug fix is still valid though - or let me know if you still wanna change all occurrences.

hhrutter: I am happy to merge this in if you limit the change to the obvious bugfix.
I am going to take care of the spelling issue later on myself.

hhrutter:
Do you think you can squash your contribution into one [signed commit](https://help.github.com/en/github/authenticating-to-github/signing-commits)?

PR #200, opened 2020-06-11
Fix test that was comparing the same variable.

minusworld:
Hey there!
I was [scanning a bunch of Go code](https://semgrep.live/scan/106a6a66-c996-4b1d-8cd0-1102f1ed4b26) and found this issue.
It looks like this test case in `pkg/pdfcpu/image_test.go` was intending to compare `bb1` with `bb2`, but it was comparing `bb1` twice.
Anyway, I hope this helps!


hhrutter:
Thanks for discovering this!

PR #208, opened 2020-07-07
Handle non *os.File buffers. (Fix #207)

### LINKED ISSUE #207: pdfcpu: missing write file descriptor
Hi!

This is breaking change https://github.com/pdfcpu/pdfcpu/commit/d8b8be9e0a1ee44fce6cb546ea0c0c129d5fca7f#diff-1f58690efd2b9d45abf9d269399235baR1031-R1034 on 0.3.4 release.  We are passing buffer based io.Writer and we have no intention to write any files.
Based on my quick analysis, it seems that we are forced to pass `*os.File`, otherwise Write will always fail on `setFileSizeOfWrittenFile`.

Please consider allowing to pass in-memory buffers that are not file associated.

As a workaround, we are reverting to 0.3.3 version.

jozuenoon:
Related to https://github.com/pdfcpu/pdfcpu/issues/207.

### DIFF
--- a/pkg/pdfcpu/context.go
+++ b/pkg/pdfcpu/context.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"fmt"
 	"io"
+	"os"
 	"sort"
 	"strings"
 
@@ -719,7 +720,7 @@ type WriteContext struct {
 
 	// The PDF-File which gets generated.
 	*bufio.Writer                     // A writer associated with Fp.
-	Fp                  io.Writer     // A file pointer needed for detecting FileSize.
+	Fp                  *os.File      // A file pointer needed for detecting FileSize.
 	FileSize            int64         // The size of the written file.
 	DirName             string        // The output directory.
 	FileName            string        // The output file name.
--- a/pkg/pdfcpu/write.go
+++ b/pkg/pdfcpu/write.go
@@ -31,9 +31,7 @@ import (
 )
 
 // Write generates a PDF file for the cross reference table contained in Context.
-func Write(ctx *Context) error {
-	var err error
-
+func Write(ctx *Context) (err error) {
 	// Create a writer for dirname and filename if not already supplied.
 	if ctx.Write.Writer == nil {
 
--- a/pkg/pdfcpu/write.go
+++ b/pkg/pdfcpu/write.go
@@ -32,8 +32,6 @@ import (
 
 // Write generates a PDF file for the cross reference table contained in Context.
 func Write(ctx *Context) error {
-
-	var file *os.File
 	var err error
 
 	// Create a writer for dirname and filename if not already supplied.
@@ -42,12 +40,13 @@ func Write(ctx *Context) error {
 		fileName := filepath.Join(ctx.Write.DirName, ctx.Write.FileName)
 		log.CLI.Printf("writing to %s\n", fileName)
 
-		file, err = os.Create(fileName)
+		file, err := os.Create(fileName)
 		if err != nil {
 			return errors.Wrapf(err, "can't create %s\n%s", fileName, err)
 		}
 
 		ctx.Write.Writer = bufio.NewWriter(file)
+		ctx.Write.Fp = file
 
 		defer func() {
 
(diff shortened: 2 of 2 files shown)


jozuenoon:
Code snippet that fails on 0.3.4:
```
package main

import (
	"bytes"
	"io"

	"github.com/mattetti/filebuffer"
	"github.com/pdfcpu/pdfcpu/pkg/api"

	"github.com/jung-kurt/gofpdf"
)

func main() {

	pdfs := []io.ReadSeeker{
		generatePDF("example 1"),
		generatePDF("example 2"),
	}

	newPDF := &bytes.Buffer{}

	if err := api.Merge(pdfs, newPDF, nil); err != nil {
		panic(err)
	}
}

func generatePDF(text string) io.ReadSeeker {
	var buf bytes.Buffer

	pdf := gofpdf.New("P", "mm", "A4", "")
	pdf.AddPage()
	pdf.SetFont("Arial", "", 14)
	pdf.Text(10, 50, text)

	if err := pdf.Output(&buf); err != nil {
		panic(err)
	}

	return filebuffer.New(buf.Bytes())
}
```

pkg/pdfcpu/context.go
```diff
@@ -719,7 +720,7 @@ type WriteContext struct {
 
 	// The PDF-File which gets generated.
 	*bufio.Writer                     // A writer associated with Fp.
-	Fp                  io.Writer     // A file pointer needed for detecting FileSize.
+	Fp                  *os.File      // A file pointer needed for detecting FileSize.
```
jozuenoon, line 723: Seems like Fp should be *os.File as the comment suggests.

pkg/pdfcpu/write.go
```diff
@@ -31,23 +31,20 @@ import (
...
 		log.CLI.Printf("writing to %s\n", fileName)
 
-		file, err = os.Create(fileName)
+		file, err := os.Create(fileName)
 		if err != nil {
 			return errors.Wrapf(err, "can't create %s\n%s", fileName, err)
 		}
 
 		ctx.Write.Writer = bufio.NewWriter(file)
+		ctx.Write.Fp = file
```
jozuenoon, line 47: I think that we can safely override Fp while Writer is nil.

pkg/pdfcpu/write.go
```diff
@@ -127,7 +124,7 @@ func Write(ctx *Context) error {
 		return err
 	}
 
-	err = setFileSizeOfWrittenFile(ctx.Write, file)
+	err = setFileSizeOfWrittenFile(ctx.Write)
```
jozuenoon, line 127: Since created Fp is inside context, we can drop the file.

pkg/pdfcpu/write.go
```diff
@@ -1017,24 +1014,18 @@ func writeXRef(ctx *Context) error {
...
 
-	f1 := f
-
 	// If writing is Writer based then f is nil.
-	if f == nil {
-		var ok bool
-		if f1, ok = w.Fp.(*os.File); !ok {
-			return errors.New("pdfcpu: missing write file descriptor")
-		}
+	if w.Fp == nil {
```
jozuenoon, line 1024: Now if we send io.Writer (non *os.File) we need to exit here.

pkg/pdfcpu/write.go
```diff
@@ -1017,24 +1014,18 @@ func writeXRef(ctx *Context) error {
...
 	if err != nil {
 		return err
 	}
 
-	f1 := f
-
 	// If writing is Writer based then f is nil.
-	if f == nil {
-		var ok bool
-		if f1, ok = w.Fp.(*os.File); !ok {
```
jozuenoon, line 1032: When io.Writer was non *os.File thw w.Fp was nil.

lgonzalez-silen:
I can confirm that this PR fixes Merge for me locally. Thanks for the PR!

hhrutter:
Good to know. 
I am still testing.

pkg/pdfcpu/write.go
hhrutter: We don't really need this, do we?

hhrutter:
Thanks for discovering this! 💚

PR #243, opened 2020-10-29
fix panic on zero index

VMorozov-wh:


hhrutter:
Hello!

I have a couple of questions:

1) Is this fixing any pdfcpu issue? If yes please provide the issue number.

2) Can you please explain this fix? What is the actual underlying problem?

3) Can you provide a sample in order to reproduce the problem?

Thank you!

VMorozov-wh:
Hello!

1. no
2. panic occurs if the pdf contains the following code:
```
3 0 obj
<</Length 2338
/Type
/Metadata
/Subtype 
/XML>>
stream
endstream
endobj
```

3. I cannot provide the original document with the problem, because it contains personal information, but I tried to simulate the problem
[example.pdf](https://github.com/pdfcpu/pdfcpu/files/5465040/example.pdf)

hhrutter (approved):

PR #247, opened 2020-11-10
sort by file name in merge pdf

AdolphKevin:
when used cmd 「merge out.pdf *.pdf 」.

merge pdf content mix. so, add this function.

### DIFF
--- a/pkg/api/merge.go
+++ b/pkg/api/merge.go
@@ -19,6 +19,7 @@ package api
 import (
 	"io"
 	"os"
+	"sort"
 	"time"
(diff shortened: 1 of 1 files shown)


hhrutter:
Hello!

I understand that sorting sounds usefull if you use *.pdf
but in other cases where I want to merge some files in a specific order we would ignore it
with the result that files are not merged in the right order.

hhrutter:
This is addressed by v0.3.8 which introduces a *sort* flag for the *merge* command.

PR #248, opened 2020-11-14
fix missing reader size

benoitkugler:
Hi, 

[This commit](https://github.com/pdfcpu/pdfcpu/commit/0b3fcc7e006f3a096866183d3ca2c12a348202a3) introduce a new check in file pdfcpu/read.go, line 199: `offset >= ctx.Read.FileSize` 
The problem is that the quantity `ctx.Read.FileSize` is only set when the `io.ReadSeeker` is a `*os.File`.
In particular, it breaks the use of in-memory `ReadSeeker`s. 

Fortunately, we can retrieve the size of a general `ReadSeeker` by seeking to the end.


hhrutter:

PR #255, opened 2020-11-30
pdfpcu/filter: Do not read after EOD

benoitkugler:
This PR slightly modifies the internal behavior of 5 filters (ASCII85, ASCIIHex, RunLength, LZW, Flate), to ensure that, given an input which implements io.ByteReader, the decoding of its data stops exactly at the EOD marker of the filter.
Such a change will be necessary to successfully parse content stream with inline image data. 

The external API is not changed, and the tests pass.

_Implementation details:_ 

LZW and Flate backing decoders already have such behavior, as indicated in their documentation: https://golang.org/pkg/compress/zlib/#NewReader says 

>  If r does not implement io.ByteReader, the decompressor may read more data than necessary from r.

and https://golang.org/pkg/compress/lzw/#NewReader

> If r does not also implement io.ByteReader, the decompressor may read more data than necessary from r.

The RunLength implementation is modified to use a `io.ByteReader`.

For ASCII85 and ASCIIHex, I implemented a "LimitedReader", very similar to the method `Reader.ReadSlice(delim byte)` provided by bufio (but bufio only supports a single byte as delimiter).
(Maybe the standard library provides a way to achieve the same result ?)

**Limitation**
pdfcpu/filter also supports decoding CCITT data, but this PR does not handle this filter.

### DIFF
--- a/pkg/filter/ascii85Decode.go
+++ b/pkg/filter/ascii85Decode.go
@@ -52,6 +52,10 @@ func (f ascii85Decode) Encode(r io.Reader) (io.Reader, error) {
 
 // Decode implements decoding for an ASCII85Decode filter.
 func (f ascii85Decode) Decode(r io.Reader) (io.Reader, error) {
+	// when possible, we make sure not to read passed EOD
+	if byteReader, ok := r.(io.ByteReader); ok {
+		r = newReacher(byteReader, []byte(eodASCII85))
+	}
 
 	p, err := ioutil.ReadAll(r)
 	if err != nil {
@@ -66,11 +70,5 @@ func (f ascii85Decode) Decode(r io.Reader) (io.Reader, error) {
 	p = p[:len(p)-2]
 
 	decoder := ascii85.NewDecoder(bytes.NewReader(p))
-
-	buf, err := ioutil.ReadAll(decoder)
-	if err != nil {
-		return nil, err
-	}
-
-	return bytes.NewBuffer(buf), nil
+	return decoder, nil
 }
--- a/pkg/filter/asciiHexDecode.go
+++ b/pkg/filter/asciiHexDecode.go
@@ -75,13 +75,5 @@ func (f asciiHexDecode) Decode(r io.Reader) (io.Reader, error) {
 		p = append(p, '0')
 	}
 
-	// dst := make([]byte, hex.DecodedLen(len(p)))
-
 	return hex.NewDecoder(bytes.NewReader(p)), nil
-	// _, err = hex.Decode(dst, p)
-	// if err != nil {
-	// 	return nil, err
-	// }
-
-	// return bytes.NewBuffer(dst), nil
 }
--- a/pkg/filter/asciiHexDecode.go
+++ b/pkg/filter/asciiHexDecode.go
@@ -48,6 +48,10 @@ func (f asciiHexDecode) Encode(r io.Reader) (io.Reader, error) {
 
 // Decode implements decoding for an ASCIIHexDecode filter.
 func (f asciiHexDecode) Decode(r io.Reader) (io.Reader, error) {
+	// when possible, we make sure not to read passed EOD
+	if byteReader, ok := r.(io.ByteReader); ok {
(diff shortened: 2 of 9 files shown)


hhrutter:
Please bear with me I am analyzing your contribution.

benoitkugler:
Hi, 

I've pushed a commit adding a sort of DCT (JPEG) decoding support. 
The idea is still to enable inline image parsing in content stream.

In this case, as you have already noticed, it would be a bit awkward to include DCT in the filter API (essentially because we dont want, given an image.Image, to write it back as an not encoded image content stream; and it would be pretty useless anyway...).

Still, we need to be able to process binary inline image content, event if we discard the decoded content. To do so, the commit provides a (kind of weird) function `LimitedDCTDecoder` which micmics the filter API.
(In the consuming package, I then special case the DCT filter to use this function.)

You could argue that this function does not belong to the pdfcpu/filter package. The only argument I see is to keep "filter-related code" in the same package.

Again, this PR is only needed by the perspective to parse content streams; so maybe it could be worth to use a separate API (maybe in a sub-package package) ?

hhrutter:
Hello!
Can you give me a minimalistic example PDF for this issue?
Thank you!

benoitkugler:
Sure !
Here is an example
[inline_images.pdf](https://github.com/pdfcpu/pdfcpu/files/5667394/inline_images.pdf)
There is one filter per page.

Note that I didn't ensure that the parsing would fail without my PR. 
Doing so require a bit more work, since one need to craft binary data that would confuse the current parsing logic. 
Do you need it as well ?

hhrutter:
Who is the PDFWriter/Creator of this?
There is no info dict.

benoit-kugler:
This file is a toy example that I generated. It is probably not realistic at all, but it stills provides a valid content.

> pdfcpu is not concerned with the actual page content of files being processed with one exception
> where we parse the structure of content streams for used resources for optimization but this does not involve decoding inline image bits.

Understood ! The issue is that, for this particular task, you have to handle inline image data, even if you don't process their content. So far, I have not encountered any trouble, but _in theory_, it could happen with the current implementation, since, to know where the data exactly ends, you have to somewhat understand the filter encoding.
Consider for instance a case where the command EI (end image) is present in the encoded binary data: just searching for EI bytes would not yield a correct result.

So, this PR is an attempt to consolidate the parsing of content streams, but do not solves a real case issue.

hhrutter:
I am kind of starting to see the problem but this PR still looks very theoretical to me.

The work in *parseContent.go* is really basic and as soon as somebody raises an issue involving it we'll take it from there and provide a fix. I can always parse a text stream in both directions - right?

Also given all possible options for data encoding via PDF filters I recall a strong recommendation for "FlateDecode" in one of the PDF spec revisions which should decrease the weight of this issue.
(Please don't put me on the spot here, because I'd need to dig that info up.)

I prefer to solve problems based on concrete issues, where I also can write a corresponding test function that proves a particular fix works as intended. 

To be really honest I am thinking twice before touching the filter package, this is critical code at the heart of pdfcpu.

A more general question would be: Do you have an idea about who uses inline images at all?
Yes, it's one of two ways to include image data in a PDF page but it comes with its limitations eg. regarding stream length.
I just have not come across many PDFs with inline images. If I were to write a PDFWriter I would try to avoid them because of its size limitations and all and just use regular object streams, but hey that's just me..

This also makes me wonder about the original intentions for providing for the possibility of inline images in the PDF spec.
Maybe there are usecases out there where this makes sense, I just hate to speculate.

benoitkugler:
> A more general question would be: Do you have an idea about who uses inline images at all?

Haha, not at all ! As you already said, this is really a theoretical PR, trying to be exhaustive and anticipate weird, but valid, PDF files.

> I prefer to solve problems based on concrete issues, where I also can write a corresponding test function that proves a particular fix works as intended.

All right, I understand ! Feel free to close the PR, then.

hhrutter:
👍  I am gonna save this for a rainy day then.
I am sure the issue will pop sometime in the future in one way or another.
I appreciate all your research into this.

PR #278, opened 2020-12-31
Add booklet option for nup 

adamgreenhall:
Hi there,

This PR adds the option to make a booklet using `nup`. Booklets are useful for making zines or other small books where the pages are put together by folding. They are basically just 2-up ([example](https://github.com/adamgreenhall/pdfcpu/blob/0f39c585d43a1cba57be331c955ea96f921e8f87/pkg/samples/nup/NUpFromPDFWithBooklet.pdf)) or 4-up ([example](https://github.com/adamgreenhall/pdfcpu/blob/0f39c585d43a1cba57be331c955ea96f921e8f87/pkg/samples/nup/NUpFromPDFWithBookletLedger.pdf)) layouts, with some page reordering to make the folding work (and rotation to make the paper cutting easier, if using the 4-up version). We originally [figured this out](https://ameliagreenhall.com/setting-up-letter-sized-zines-to-print-on-riso/) using [a latex package](https://ctan.org/tex-archive/macros/latex/contrib/pdfpages) - but wanted to make it work in `pdfcpu`.

Happy to change things around if you have suggestions.


hhrutter:
Hello!

I think this is a nice usecase that deserves its own pdfpcu operation 💚 

How does something like the following sound:
```
pdfcpu booklet description in.pdf out.pdf
```

OK that's assuming I understand this correctly.
This essentially boils down to a nup From PDF selecting all pages with some clever rotation applied to a custom page sequence?

Please fill me in on the details of this command and (all) needed arguments that make sense.
We need to think about if this command should follow the cli syntax of nup or if we could make it even simpler.

Also how would you document this operation in one paragraph?

As far as samples *Walden* has a lot of pages so the demo effect for this operation is limited I think.
Some nice sample(s) that use(s) a small set of pages so potential users get a better big picture of this command would be a great addition to go along. Maybe something where the folding idea is transported in an intuitive way 🤔 

Ideally I don't want to mess around too much in pdfcpu.nup.go and I am also not in favour of adding this on top of the current nup command via the booklet orientation.

Yes this does concern orientation, but it is actually much more because of a custom page sequence
so maybe there is a better a way. Maybe a pdfcpu.booklet.go but then we want to avoid code repitition..
I don't know what do you think?

I hope I am not 🙏  too much. 
In case there is a lack of time on your side I am happy to craft a proper command and integrate that.
In this case I'd still need some more info as stated above.

adamgreenhall:
All of those suggestions sound good. Will start on that next week.

adamgreenhall:
Made a shorter and simpler example than Walden 😬 : [booklet 2-up](https://github.com/pdfcpu/pdfcpu/blob/3d59e0f70a213bcc36e3b68e9de8c60a4e58f5e0/pkg/samples/nup/NUpFromPDFWithBooklet.pdf) and [booklet 4-up](https://github.com/pdfcpu/pdfcpu/blob/3d59e0f70a213bcc36e3b68e9de8c60a4e58f5e0/pkg/samples/nup/NUpFromPDFWithBookletLedger.pdf).

Going to start pulling out the overrides to nup and moving them to their own booklet cmd/api.

Does the following make sense to you as a description?

`booklet` arranges a sequence of pages onto larger sheets of paper. Using this arrangement, you can assemble a small book or zine. There are two styles of booklet, depending on your paper/input and sheet/output size.  Here are the assembly instructions:
* If two of your pages fit on one side of a sheet (eg statement on letter or A5 on A4), assemble by printing on both sides (odd pages on the front and even pages on the back) and folding down the middle:
Zinemaking-folding-folio1
Zinemaking-folding-folio4

* If four of your pages fit on one side of a sheet (eg statement on ledger or A5 on A3), assemble by printing on both sides, then cutting the sheets horizontally. The sets of pages on the bottom of the sheet are rotated so that the cut side of the paper is on the bottom of the booklet for every page. After cutting, place the bottom set of pages after the top set of pages in the booklet. Then fold. Sort of like this (but you cut horizontally instead of fold):
Zinemaking-folding-quarto


(images from https://en.m.wikibooks.org/wiki/Zine_Making/Putting_pages_together)

hhrutter:
Yeah, nice! 👍 

The description sounds good to me.
We'll need a short and a long one for the cli usage. (usage.go)
As far as pdfcpu.io I am not sure I want to reuse the images from wikibooks but that's not smth we need right away.

adamgreenhall:
Added `booklet` to the CLI and moved the booklet related logic out of nup. There's some duplication between nup and booklet - but hopefully not too bad. Let me know what you think.

hhrutter:
Thanks!  👍 
I'll check it out

hhrutter:
So booklet-letter.pdf is a bit confusing to me.
How am I supposed to fold this? 
Could it be that here is smth upside down?

pkg/cli/cmd.go
```diff
@@ -39,6 +39,7 @@ type Command struct {
 	Import         *pdfcpu.Import
 	Rotation       int
 	NUp            *pdfcpu.NUp
+	Booklet        *pdfcpu.Booklet
```
hhrutter, line 42: Get rid of Booklet.
Use Nup also for booklets.

pkg/api/booklet.go
```diff
@@ -0,0 +1,108 @@
...
+	"io"
+	"os"
+	"time"
+
+	"github.com/pdfcpu/pdfcpu/pkg/log"
+	"github.com/pdfcpu/pdfcpu/pkg/pdfcpu"
+)
+
+// PDFBooklet rearranges PDF pages or images into a booklet layout and writes the result to outFile.
+func PDFBooklet(inFiles []string, outFile string, selectedPages []string, booklet *pdfcpu.Booklet, conf *pdfcpu.Configuration) (err error) {
```
hhrutter, line 29: Rename to BookletFile for consistency

pkg/pdfcpu/booklet.go
```diff
@@ -0,0 +1,314 @@
...
+}
+
+// BookletFromPDF creates an booklet version of the PDF represented by xRefTable.
+func (ctx *Context) BookletFromPDF(selectedPages IntSet, booklet *Booklet) error {
+	grid := &Dim{
+		math.Round(booklet.SheetDim.Width / booklet.PageDim.Width),
+		math.Round(booklet.SheetDim.Height / booklet.PageDim.Height),
+	}
+	n := int(grid.Width * grid.Height)
+	if !(n == 2 || n == 4) {
```
hhrutter, line 151: I think we should redesign the whole command to the way nup works.
`pdfcpu booklet [-p(ages) selectedPages] [description] outFile n inFile|imageFiles...``

Rather than asking the user to provide the input and output page size,
let him decide between 2up or 4up booklet (n= 2 or 4) and require a sheet page size (the size of the paper that gets printed). The user has to understand what that means so e have to explain it in the usage correspondingly. This way we can also reuse the Nup struct.

This also makes sense if you want to support booklets based on image files.

hhrutter:
pdfcpu/nup.go: calcTransMatrixForRect

This modification does not work.

First of all the matrix transformation has to be the well defined sequence if:
1) Scale
2) Rotate
3) Translate

Then you are adding code that does not have any effect when you modify m2 after calculating m:

```
m := m1.multiply(m2).multiply(m3)

if rotatePageForBooklet {
	// we've rotated 180deg, so we need to translate to get the old page visiible on the new page
	if rot == 0 {
		m[2][0] += w * r1.Width()
		m[2][1] += h * r1.Height()
	} else {
		// 90 degree rotation because of paper, in addition to the 180 rotation for booklet
		// need to shift rotated page on sheet down and left
		m[2][0] -= h * r1.Width()
		m[2][1] += w * r1.Height()
	}
}

return m
```


It would be good if you could provide the modified version of this for booklet within booklet.go for now.
In other words please leave pdfcpu/nup.go untouched for now.

adamgreenhall:
> So booklet-letter.pdf is a bit confusing to me.
> How am I supposed to fold this?
> Could it be that here is smth upside down?

Just started at this a bit more today - there was a page ordering bug in the 2-up version. Thanks for checking. The folding is done horizontally (along the shorter 8.5" dimension of the letter paper). This makes a statement sized booklet that has a fold on the longer dimension (the 8.5" side). Just printed it out and folded to double check.

pkg/pdfcpu/booklet.go
adamgreenhall: Makes sense - will restructure a bit.

adamgreenhall:
@hhrutter cleaned things up (transition matrix, configs, and usage). There're are still a few edits in `nup.go`, but hopefully ok now. Let me know what you think.

hhrutter:
I know you are putting in a lot of valuable time and I appreciate it but I have to ask you for a favor:
Other than using the nup struct do you think you could keep the booklet code cleanly separated from nup?
I want to take care of the integration myself which is subtle because of code base consistency and readability concerns.
Thank you!

hhrutter:
This PR has been merged in and is the main feature of v0.3.9 (which is still in the cutting room)
Adam, thank you so much for your contribution 💚 
Highly appreciated 👍

PR #283, opened 2021-01-22
Recursively retrieve the entire outline structure

JanBaryla:
Hey!

My use case required getting access to the entire bookmark structure. I was processing CAD-exported architectural plans, and the meaningful names of those bookmarks were really important.

As it turned out, pdfcpu only handled the first level of bookmarks, so I took the liberty of digging into what can be done to retrieve the entire thing.

Moreover, I noticed the particular PDF I was parsing didn't make use of the `Dest` field of the Outline. Instead, it defined a `GoTo` type `A` field, which pretty much acted the same way. So I added support for this too.

I'm gonna rely on my fork for the time being, but I thought it would be pretty cool to make this an official feature. I'll gladly listen to any suggestions or change requests.


hhrutter:
💚  Thanks for putting time into this.
I'll check it out and will feed back.

hhrutter:
Sorry it's only now I can tackle this..

There is a reason why the split by bookmark feature is limited to the first level.
In general bookmarks corresponding to some logical chapter or section in your PDF don't align with the top of a new page.
So there is always the issue of which pages to render for a specific bookmark.
You only get the first page via the *Dest* of a specific *bookmark*.

For example if you have 3 *bookmarks* on one page and on top the third runs over to the next page which pages are you gonna include in these 3 *bookmark* files?

So for the most part the whole thing is not very intuitive and I am not even sure whether to keep it at all because of this.
And it gets more complicated when you want to render multiple levels of bookmarks to corresponding files.
What are your rules for when level n starts on top of page 5 and level n+1 also starts on page 5 just a few lines below?
I only see an opinionated approach to this problem and that might not be useful for general user of this command.

Only and if all of your bookmarks no matter at what level in the bookmark hierarchy point to the top of a new page this operation maybe useful.

After all I am curious now how the bookmarks are organized in your case.

PS: Extending the operation to consider *Goto Actions* is an improvement that makes sense.

JanBaryla:
Hey, thanks for the answer.

> There is a reason why the split by bookmark feature is limited to the first level.

Yes, and if I'm not mistaken, I did not change that behavior. The bookmark structure retrieved by my code has more data in it, but that splitting feature still only goes through the first layer.

> For example if you have 3 bookmarks on one page and on top the third runs over to the next page which pages are you gonna include in these 3 bookmark files?

This is, indeed, very tricky. I have personally not seen use cases like this, and I wonder how they'd be handled by professional PDF browsers. My code, however, doesn't really change what you had before. The way the original `BookmarksForOutlineLevel1` implemented figuring out the `PageThru` value was naive, and I preserved that - `PageFrom` is retrieved from the PDF structure and `PageThru` may have a value, or maybe be left unintialized if it's the end of the bookmark. What I'm trying to say is - I'm hesitant to admit my code complicates things. It follows the pattern introduced by the original - squeezing more data out of the underlying PDF structure.

> I only see an opinionated approach to this problem and that might not be useful for general user of this command.

I only wrote this code because pdfcpu is great but I really needed to get the nested bookmark data for my use case. I did not really intend for it to become a command available to the user - rather just a part of the API for any people who might have similar use cases. I of course see your point, where you'd like your tool to be as useful to end-users as possible. I'm open to either modifying the code or not merging this at all. But PDF browsers do support nested bookmarks, so I bet this can be pulled off.

> Do you think you can share a sample PDF?

I can't disclose it publicly but I sent it to your email.

> PS: Extending the operation to consider Goto Actions is an improvement that makes sense.

Let's see how testing on that PDF goes and where this discussion leads us. If we end up agreeing that recursive bookmarks are not something worth pursuing officially, I can at least remake this improvement into its own PR.

hhrutter:
I am gonna run some tests and report back.

hhrutter:
I did not realize this PR was on the API only.
LGTM 👍

PR #321, opened 2021-04-12
Extract images in memory

rgargente:
Hello, 

We were doing some experiments last friday with your library. We needed to extract images but didn't want to work with files so I made some modifications to the library to have a function that allows to work with images in memory, basically skipping the writing file parts.

I didn't modify the original interface at all so all existing code should work just as before.

I also added some testing code.

Greetings from the EndNote Click team.

### DIFF
--- a/pkg/api/extract.go
+++ b/pkg/api/extract.go
@@ -36,11 +36,13 @@ func ExtractImages(rs io.ReadSeeker, outDir, fileName string, selectedPages []st
 	return error
 }
 
+// Similar to ExtractImages but returnes in memory images as opposed to writing them to disk.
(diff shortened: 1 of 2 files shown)


hhrutter:
I refactored this into
`func ExtractImagesRaw(rs io.ReadSeeker, selectedPages []string, conf *pdfcpu.Configuration) ([]pdfcpu.Image, error) `

which is called by 
`func ExtractImages(rs io.ReadSeeker, outDir, fileName string, selectedPages []string, conf *pdfcpu.Configuration) error `

which is called by 
`func ExtractImagesFile(inFile, outDir string, selectedPages []string, conf *pdfcpu.Configuration) error`

PR #328, opened 2021-05-03
Fix rename err

omer-akbas:
Cannot change filename because the file is open (f1) and gives rename error. It must be closed before changing its name.


hhrutter:
Hello!

Please explain how this issue can be reproduced.
Thank you!

hhrutter:
Thanks for discovering this!
This seems to be an issue on the Windows platform only.
I activated the corresponding tests which were failing on Win for some reason and now we know the reason..

PR #334, opened 2021-05-20
Fix error invalid encoding SymbolSetEncoding

truyet:
I have a pdf file run extract with `error validateFontEncoding: invalid Encoding name: SymbolSetEncoding` so we need update font encoding
[test1.pdf](https://github.com/pdfcpu/pdfcpu/files/6512974/test1.pdf)


hhrutter:
Hello, I am unable to accept this unless you sign the CLA.

PR #336, opened 2021-05-26
Fix index-out-of-range error for PDF with corrupted dictionary

christiannicola:
Hi!

I'm encountering an issue in production with the latest release - 'index-out-of-range' error while processing a PDF with a seemingly corrupted document info dictionary.

I'm unable to share the PDF itself, but I can share a gist containing the `pdfcpu validate -vv` output (with some small redactions).

https://gist.github.com/christiannicola/f7757ed69cfe684f02ea4a774ad2b08f 

This PR fixes the runtime panic, and the method properly returns an error stating that the dictionary is corrupt - I'm assuming that that was the intention anyway.

Thank you for your consideration and this awesome library!

### DIFF
--- a/pkg/pdfcpu/read.go
+++ b/pkg/pdfcpu/read.go
@@ -1479,7 +1479,7 @@ func buildFilterPipeline(ctx *Context, filterArray, decodeParmsArr Array, decode
 		if !ok {
 			return nil, errors.New("pdfcpu: buildFilterPipeline: filterArray elements corrupt")
 		}
-		if decodeParms == nil || decodeParmsArr[i] == nil {
+		if decodeParms == nil || (len(filterArray) <= len(decodeParmsArr) && decodeParmsArr[i] == nil) {
 			filterPipeline = append(filterPipeline, PDFFilter{Name: filterName.Value(), DecodeParms: nil})
 			continue
 		}


hhrutter:
👍 I need to check this.

hhrutter:
This code assumes filterArray and decodeParmsArr have the same length which does not necessarily have to be that way obviously.

christiannicola:
This was only an attempt to safeguard against the error. The index `i` comes from the `for range` over `filterArray` loop, so we should assert that `decodedParams` is at least of `len(filterArray)`.

Not very elegant, but then again I only spent about 30 minutes with the source code.  I'm happy to rework this if you have suggestions!

hhrutter:
It's a little more complicated.
This looks more like a parsing problem.
The length of both arrays should be the same.

The spec says:
_...If there are multiple filters and any of the filters has parameters set to nondefault values, DecodeParms shall be an array with one entry for each filter: either the parameter dictionary for that filter, or the null object if that filter has no parameters (or if all of its parameters have their default values)._ 

You have a temp fix for this which is good.
I am in the process of providing support for cascaded filters anyway across the board and this will also correct filter pipelines.

hhrutter:
This is resolved with v0.3.12

PR #346, opened 2021-06-24
Watermark remove white objects

mipmip:
To make watermark visiable on non-transparent document i have a method to remove all white rectangles.

I've have working example removing substrings of the content stream. Everything between "/Cs1 cs 1 1 1 sc" and "/Cs1 cs 0 0 0 sc" and everything between "1 1 1 rg", "0 0 0 rg" is removed.

This might be a little simplistic, but I it used to work quiet fine in a Mac app I made. I wonder if this could be part of pdfcpu or if you have comments on this.

### DIFF
--- a/pkg/api/extract.go
+++ b/pkg/api/extract.go
@@ -270,6 +270,7 @@ func ExtractContent(rs io.ReadSeeker, outDir, fileName string, selectedPages []s
 		if r == nil {
 			continue
 		}
+
 		outFile := filepath.Join(outDir, fmt.Sprintf("%s_Content_page_%d.txt", fileName, p))
 		log.CLI.Printf("writing %s\n", outFile)
 		f, err := os.Create(outFile)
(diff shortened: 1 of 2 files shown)


hhrutter:
Hi, thanks for your contribution.

Could you maybe share an example, I don't fully understand the issue here.

Stamps get drawn as the last content whereas watermarks get drawn first.
If a watermark or some part of it is hidden in general my recommendation is to use a stamp with some opaqueness.
That usually serves the purpose.

hhrutter:
In general I think patching content streams like you are suggesting is smth too invasive for pdfcpu.

mipmip:
Here are two pdf which can be used to test this feature.

[watermark.pdf](https://github.com/pdfcpu/pdfcpu/files/6732182/achtergrond.pdf)
[letter-non-transparent.pdf](https://github.com/pdfcpu/pdfcpu/files/6732187/tekstdocument.pdf)

The letter was made with Apple Pages.

I run  ```./pdfcpu watermark add -m pdf -- "achtergrond.pdf" "rot:0, scale:1.0 rel" tekstdocument.pdf out1.pdf```

hhrutter:
Sorry - I don't feel confident making this kind of content stream manipulation part of pdfcpu.

PR #360, opened 2021-07-24
Allow for verbose parameters

signalwerk:
I think a command like `al:j` works great if you are very familiar with the command. But if you wanna have others on board you probably use `aligntext: j`. But it might be greate if you can also use `aligntext: justify`. I addes some more alternatives

PR #361, opened 2021-07-24
add documentation for verbose parameters and unify syntax (stamp and watermark)

signalwerk:
I tried to unify the style and notation of the parameters.

The logic for the values column is now:

`{values} <br> {explanation}, {examples}`

* `{values}` this follows now a logic with; `or|combination`, `[optional]`, `{variable}`, `fixtext`
* `{explanation}` explanation with parameters
* `{examples}`

### DIFF
--- a/core/stamp.md
+++ b/core/stamp.md
@@ -115,19 +115,20 @@ but `op: .7` will do the job.
 | fontname                    | a basefont                                        | `{name}` <br> Please refer to `pdfcpu fonts list`                                                                                                                                                            | `font:Helvetica`
 | points                      | fontsize in points                                | `{size}` <br> in combination with absolute scaling only                                                                                                                                                      | `points:24`
 | rtl                         | right to left userfont                            | `{rtl}` <br> one of: `on|true|t`, `off|false|f`                                                                                                                                                              | `rtl:off`
-| position                    | the stamps lower left corner                      | `{position}` <br> one of: `f|full` or anchor `tl|top-left`, `tc|top-center`, `tr|top-right`, `l|left`, `c|center`, `r|right`, `bl|bottom-left`, `bc|bottom-center`, `br|bottom-right`                        | `pos:c`
-| offset                      |                                                   | `{dx} {dy}` <br> in user units  <br> eg. `15 20`                                                                                                                                                             | `off:0 0`
-| scalefactor                 |                                                   | `{factor} [mode]` <br> 0.0 < `{factor}` <= 1.0, optional `{mode}` one of: `abs|absolute`, `rel|relative`                                                                                                     | `sc:0.5 rel`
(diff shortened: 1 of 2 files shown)


signalwerk:
Please keep in mind I tested the markup only with [stackedit.io](https://stackedit.io/), please have a look at the generated HTML

signalwerk:
I made further adjustments to the doc after the update of usage. Both are in sync now.

hhrutter:
It turns out the `pdfcpu boxes` command is also using the positioning anchors.
My bad :(
Since the doc is already getting messy I am gonna spawn out a separate section for positioning anchors.
I will update the source code and docs accordingly when I push the next release.

signalwerk:
@hhrutter just to understand why the PR is closed; 
you say it's better not to have the current documentation updated at all because there is something missing in an other part? Or what exactly is the impediment?

I tried to fix many inconsistencies inside the doc and I'm not sure what hold you back.

hhrutter:
Like I said, there are three commands that deal with positioning anchors: `import`, `stamp` and `boxes`.
`boxes` slipped through unfortunately.

In order to be consistent we need to adapt the verbose style to all three commands in the code base and in the documentation. Since the boxes command is quite complex and the doc is even more complex I need to postpone this until I have time for it including some rearranging because it is getting hard to maintain.

It would have been a good idea if we could have discussed all "inconsistencies" by you opening an issue before you started putting your time into a PR. The idea here is to get the user community involved and get to some consensus.
Generally I prefer to have a short discussion with users before they come up with PRs. Often times there are assumptions about certain things and misunderstandings that can get resolved easily in eg Slack and after all this project is opinionated. 

Another point is that PR #360 not only took care of the verbosity of positioning anchors and alignment it also diverged and did some additional stuff not initially discussed changing things around and deleted some sections of the usage. I made the corresponding corrections and had to get rid of some modifications we did not talk about before merging that in.
I then glanced over your modifications of the doc and noticed the same..

I am sure some of this is justified while other stuff is simply wrong possibly because you are new to the project. For example there is a difference between pdfcpu "core fonts" and "user fonts".

After all I acknowledge some inconsistencies in the docs and I will take care of them as soon as I get to it. The CLI usage and the documentation need to be consistent reflect each other and the current state of the codebase.

🙏  I appreciate your motivation and engagement but this is something I need to take care of myself - also because it means restructuring the markdown pages to some extent.

If you have any questions you can always get in touch with me on the #pdfcpu Slack channel.

signalwerk:
Apologies for the delayed response. Holiday season here.

Sure, I understand what you say. And you are probably right; at an early stage of a project it's good to have a strong authorship and I want to avoid holding you back with community work. Sometimes you just have to do stuff and not talk about it. 

On my to-do list were 2–3 other things I had in mind with pdfcpu and I try to document them to an Issue so you can then decide if you like the ideas and have time for it at one point. 

Thank you for investing your time and energy in such a great project!

PR #364, opened 2021-07-30
floating point fontsize

signalwerk:
I need to stamp a pdf with a font-size of 7.5pt. The Font-Size is just int and not float. This MR is about to change that and change some functions to float.

```sh
pdfcpu stamp add -unit 'mm' -pages '1' -mode text "Hello World" 'scalefactor: 1.0 abs, rot: 0, points: 7.5' TestNeu.pdf out.pdf
```

Please keep in mind; these are my first Go-Commits. I'm thankful if you have inputs or corrections for me.

((speaking of it: how do I run the tests? `go test` gives me `github.com/pdfcpu/pdfcpu/cmd/pdfcpu	[no test files]`))


hhrutter:
Please layout your usecase for discussion.

signalwerk:
My use case is – as described – that I have to have a font-size that has to be 7.5pt. This is exactly the font-size used in the pdf that the stamp is used. 7pt or 8pt is too much off from the original, because in small sizes 0.5pz is visual significant. It has to match the font-size of the rest of the pdf. 

((There is also the problem that I need to match the same line-height but this isssue is only about the size first))

hhrutter:
I still don't understand you high level usecase.
What is the problem your are trying to solve?
Please explain in whatever detail you need to go.

A font size of 1 points equals 1 user unit which normally equals 1/72 of an inch and then you scale up.
As far as pdfcpu stamping you can play around with font size, scaling factor and other parameters.

signalwerk:
These are two screenshots produced with the command:

```sh
pdfcpu stamp add -unit 'mm' -pages '1' -mode text "Hello World" 'scalefactor: 1.0 abs, rot: 0, points: 7' TestNeu.pdf out.pdf
```


Output:

Screenshot 2021-07-30 at 21 19 38


Detail that shows the problem:

Screenshot 2021-07-30 at 21 19 58



The stamp should have the fontsize `7.5` but for that I made the code changes.

hhrutter:
Please get in touch on Slack so we can discuss this.

signalwerk:
@hhrutter I see that this is probably not the right focus now for the project. Since it's published here, people can still pull it if they have similar cases. 

see also #361

PR #397, opened 2021-11-30
Feature: Split in buffer array in SplitBuffer

juaismar:
Like as
`func SplitFile(inFile, outDir string, span int, conf *pdfcpu.Configuration) error`,
the function 
`func SplitBuffer(rs io.ReadSeeker, span int, conf *pdfcpu.Configuration) (buffers []bytes.Buffer, err error)` 
split a file, but return files in array of buffers.

This feature prevents write in disk if you dont need store files.

Signed-off-by: Juan Iscar <juan.iscar@exponentiateam.com>

### DIFF
--- a/pkg/api/split.go
+++ b/pkg/api/split.go
@@ -17,6 +17,8 @@
 package api
 
 import (
+	"bufio"
+	"bytes"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


hhrutter:
Thanks for putting time into this.
I am in the middle of pushing a release.
Will check it out later!

hhrutter:
Hi!

Analysing your proposed changes I realized what we really need here is to return io.Readers.
While I was at it I created a PageSpan including meta info, refactored split.go for consistency and
renamed your function to SplitRaw:

```
type PageSpan struct {
	From   int
	Thru   int
	Reader io.Reader
}

func SplitRaw(rs io.ReadSeeker, span int, conf *pdfcpu.Configuration) ([]*PageSpan, error)
```

So I did not merge in anything but ss your idea is valuable to the project I added you to the contributors.

Thank your for using pdfcpu 💚

PR #411, opened 2021-12-20
Fix scaling of 1bit grayscale pixels in renderDeviceGrayToPNG

ungerik:
In `renderDeviceGrayToPNG` `color.Gray.Y` is expected to be in the range of 0 .. 255, but the default was `colValRange{0, 1}`.

I have also removed the unnecessary allocation of a `[]colValRange` slice per pixel which should improve performance.


hhrutter:
There are 2 separate steps involved here:
1) Decoding the pixel value into the source color space respecting the optional DecodeArray
2) Scaling up to 8 bpc

PR #414, opened 2021-12-23
Search PDF header in first 8k bytes

ungerik:
Improvement for function `headerVersion`:

Searching for the PDF header in the first 100 bytes was not sufficient for many files found in the wild.

Also the check for `len(s) < 8` did not work because the length of `s` was always 100 bytes.

### DIFF
--- a/pkg/pdfcpu/read.go
+++ b/pkg/pdfcpu/read.go
@@ -58,6 +58,7 @@ func ReadFile(inFile string, conf *Configuration) (*Context, error) {
 
 // Read takes a readSeeker and generates a Context,
 // an in-memory representation containing a cross reference table.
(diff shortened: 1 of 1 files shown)


hhrutter:
This has been addressed by adding  an option to config.yml:
```
# buffer size >= 100 for locating PDF header
headerBufSize: 100
```
Either keep on working with your config.yml and pdfcpu will use 100 for the headerBufSize or
1) Lookup the path of your config.yml via `pdfcpu config`
2) Backup config.yml
3) Remove config.yml
4) Issue `pdfcpu config` or any other command and pdfcpu will recreate confg.yml including the new option
5) Optionally patch *headerBufSize* to some value >= 100 as needed
6) Don't forget to restore your custom config options as well

PR #433, opened 2022-02-11
fix issue on alphanumeric sorting on filenames

mrtrkmn:
Signed-off-by: mrtrkmn <mr.turkmen@icloud.com>

When I use the pdfcpu tool I have realized an issue regarding to sorting files when we want to merge. 

The problem occurs when we have filenames which are like  

- a-2-b
- a-32-b
- a-12-a
- a-4-b

Small example to the problem can be found here: 

https://play.golang.com/p/3FuTcjOpU3j

> Expected result: [a-2-b a-4-b a-12-a a-32-b]

> Actual result: [a-12-a a-2-b a-32-b a-4-b]

### DIFF
--- a/cmd/pdfcpu/process.go
+++ b/cmd/pdfcpu/process.go
@@ -35,6 +35,8 @@ import (
 
 var errInvalidBookletID = errors.New("pdfcpu: booklet: n: one of 2, 4")
 
+const maxByte = 1<<8 - 1
+
 func hasPDFExtension(filename string) bool {
 	return strings.HasSuffix(strings.ToLower(filename), ".pdf")
 }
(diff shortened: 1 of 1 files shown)


mrtrkmn:
There is a bug that on this feature request, therefore, converted into draft.

hhrutter:
Hi there and sorry for the delay!

Your suggested file sorting extension prior to merging maskes sense unfortunately this is not something I can accept right away and here is why:

This is a CLI extension and we need to be careful to remain backwards compatible.
There is aleady a *sort* flag on the merge command and this needs to be extended.
Rest assured I will consider this and include it in the next release.

hhrutter:
For backwards compatibility reasons I have decidee the `sort` option on the CLI remains as is.
Any sophisticated sorting can easily be injected when using the API.

PR #452, opened 2022-04-11
Enhance PageCount to respect ValidationNone configuration

maxiride:
Currently, PageCount ignores the configuration parameter ValidationNone.

The condition check has been implemented following the identical use case found in other API methods.

### DIFF
--- a/pkg/api/pages.go
+++ b/pkg/api/pages.go
@@ -206,8 +206,11 @@ func PageCount(rs io.ReadSeeker, conf *pdfcpu.Configuration) (int, error) {
 	if err != nil {
 		return 0, err
 	}
-	if err := ValidateContext(ctx); err != nil {
-		return 0, err
+
+	if conf.ValidationMode != pdfcpu.ValidationNone {
+		if err = ValidateContext(ctx); err != nil {
+			return 0, err
+		}
 	}
 	return ctx.PageCount, nil
 }


hhrutter:
Hello!
As weird as it may sound but the pageCount gets computed during validation.
Se line 141: // ctx.Pagecount gets set during validation.

maxiride:
Oh, my bad, I didn't dig deep enough and I got tricked by seeing all the tests being passed. Indeed, nopw that I notice the tests for the page count are only for the PageCountFile method which provides a default configuration to the underlying PageCount method, basically bypassing the change I made altogether.

> See line 141: // ctx.Pagecount gets set during validation.

If so, is there a way to set the page count without validation? 
In real life, I'm encountering many PDF files that do not pass even the relaxed validation but are still readable, printable and with no real impact on usability. I basically have to set ValidationNone everywhere in the API method calls and not being able to have a page count is hindering the usability of the library.

hhrutter:
Since counting the pages amounts to walking down the page tree and validation has to do that anyway
right now that's where this happens. 

Also the basic idea is that any PDF passes validation, if not that spells trouble generally.
So if you have a PDF that does not pass validatio, please come forward and we extend relaxed validation
but I cannot rule out a parser bug but the point is that it would harden the pdfcpu parser,

PR #477, opened 2022-05-23
fix pararm for usage of AddTextWatermarksFile

ttrasn:
The first param of `AddTextWatermarksFile` need a source file, but it used outFile, I changed it to inFile


hhrutter:
This is intentional.
When repeatedly adding stamps to some inputFile `outFile` becomes the `inFile` of the next operation.

PR #481, opened 2022-06-01
fix buffer handling

HeavyHorst:
The code assumed that all read operations (io.Reader) are filling the given buffer completely until EOF is reached.
This is not the case with all io.Readers.

I added a helper function fillBuffer that reads from the reader until the buffer is full or EOF is reached.
Unlike io.ReadFull this function doesn't return ErrUnexpectedEOF if an EOF happens before the buffer is filled.
It behaves as a normal io.Read call but reads until the buffer is full.

### DIFF
--- a/pkg/pdfcpu/read.go
+++ b/pkg/pdfcpu/read.go
@@ -94,6 +94,25 @@ func Read(rs io.ReadSeeker, conf *Configuration) (*Context, error) {
 	return ctx, nil
 }
 
+// fillBuffer reads from the reader until buf is full or an error happens.
+// The error is EOF only if no bytes were read.
+// Unlike io.ReadFull this function doesn't return ErrUnexpectedEOF if an EOF happens before the buffer is filled.
+func fillBuffer(r io.Reader, buf []byte) (int, error) {
+	var n int
+	var err error
+	for n < len(buf) && err == nil {
+		var nn int
+		nn, err = r.Read(buf[n:])
+		n += nn
+	}
+
+	if n > 0 && err == io.EOF {
+		return n, nil
+	}
+
+	return n, err
+}
+
 // ScanLines is a split function for a Scanner that returns each line of
 // text, stripped of any trailing end-of-line marker. The returned line may
 // be empty. The end-of-line marker is one carriage return followed
@@ -213,7 +232,7 @@ func offsetLastXRefSection(ctx *Context, skip int64) (*int64, error) {
 
 		curBuf := make([]byte, bufSize)
 
-		_, err = rs.Read(curBuf)
+		_, err = fillBuffer(rs, curBuf)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


HeavyHorst:
Hallo auch!

I've added a test to demonstrate the issue.
Without this PR I get this error:
//read_test.go:52: couldn't create pdf context:  Read: xRefTable failed: pdfcpu: can't find last xref section

hhrutter:
What I need to see is the usecase / PDF that demonstrates and reproduces the issue you'd like to address with this.

HeavyHorst:
My usecase: i have an http.Filesystem that serves Files directly from ZIP Files. I Split the Files into Pages/chunks of 4MB each to Support seeking (for example in Videos). Without chunkig you have to reread everything from the beginning of the io.reader.

When the Buffer ist bigger then the current Page i read till the end of the page and Return the Bytes read. This is conform to the io.Reader Interface and should Work.

I'll try to construct a better example.

HeavyHorst:
From the Io.Reader documentation: 

Read reads up to len(p) bytes into p. It returns the number of bytes read (0 <= n <= len(p)) and any error encountered. Even if Read returns n < len(p), it may use all of p as scratch space during the call. If some data is available but not len(p) bytes, Read conventionally returns what is available instead of waiting for more.

So some totally fine Reader Implementations arent working at the moment. This PR fixes this.

hhrutter:
This may be a dumb question but where is the PDF usecase here?

HeavyHorst:
I have an http handler that returns the page Labels for a given PDF. The handler has Access to the http Filesystem. I wasnt able to give the http.file to pdfcpu directly because of this issue. Reading the whole File into a Byte Buffer worked but i searched for the underlying issue and thought i fix it.

hhrutter:
I need to know more about this..
Can we continue on the #pdfcpu Slack channel?

hhrutter:
I think I understand your usecase.

In my opinion the filling of the read buffer is the responsibility of the reader.

If you fill up the chunked read buffer in your reader (or one of your stacked readers)
like so then we don't have to change anything in read.go:

```
type smallbufferreader struct {
	rs io.ReadSeeker
}

func (sm *smallbufferreader) ReadChunk(p []byte, size int) (int, error) {
	b := make([]byte, size)
	n, err := sm.rs.Read(b)
	copy(p, b)
	return n, err
}

// Read reads up to chunkSize bytes into p. It returns the number of bytes read (0 <= n <= chunkSize) and any error encountered.
func (sm *smallbufferreader) Read(p []byte) (int, error) {
	chunkSize := 10
	var n int
	var err error

	for n < len(p) && err == nil {
		var nn int
		want := chunkSize
		if len(p)-n < chunkSize {
			want = len(p) - n
		}
		nn, err = sm.ReadChunk(p[n:], want)
		n += nn
	}

	return n, err
}

func (sm *smallbufferreader) Seek(offset int64, whence int) (int64, error) {
	return sm.rs.Seek(offset, whence)
}
```

PR #484, opened 2022-06-02
Fix JS build

scoiatael:
Hey,

looks like WASM version stopped building some time ago.

```
$ env GOOS=js GOARCH=wasm go build
# github.com/pdfcpu/pdfcpu/pkg/pdfcpu
../../pkg/pdfcpu/parseConfig_js.go:219:28: too many arguments in call to handleHeaderBufSize
        have (string, string, *Configuration)
        want (string, *Configuration)
```

This fixes the problem. Maybe WASM build should be added to GH actions since it has some build-specific files?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,4 +32,4 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Go build
-        run: env CGO=0 GOOS=${{ matrix.target.goos }} GOARCH=${{ matrix.target.goarch }} go build ./cmd/pdfcpu
+        run: env CGO_ENABLED=0 GOOS=${{ matrix.target.goos }} GOARCH=${{ matrix.target.goarch }} go build ./cmd/pdfcpu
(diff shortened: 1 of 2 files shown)


hhrutter:
Thanks for discovering this.
A separate Github action does make sense.

hhrutter:
If you can add it that would be nice.
There is already an OS matrix in test.yml and this does not fit in.
I guess we need another workflow then triggered by push, pull_request.

scoiatael:
Done - new workflow should cross-compile from Linux to a couple of standard targets. Looks like it needs approval before it can run, so kinda hard to verify if it works by myself.

hhrutter:
Thanks, this will go live together with the upcoming release.

hhrutter:
Fixed as of v0.4.0

scoiatael:
> Fixed as of v0.4.0

Looks like it was closed, but not merged? I see that the original issue seems to be fixed in 0.4.0, but the Github workflow is not there.

hhrutter:
Your contribution was integrated into the main workflow.
Had a lot of headache going on with refactoring..
Hope that's ok with you.
🙏 Thank you!

PR #502, opened 2022-08-02
Check for nil before writing to file

j0hax:
While using the pdfcpu API to extract images from a database of documents, my program would periodically crash with a `SIGSEGV`.

Although I cannot share the exact document I was processing, I was able to reproduce this on my workstation CLI using the same document:

```console
$ pdfcpu extract -mode image complete_diss.pdf .
extracting images from complete_diss.pdf into ./ ...
writing complete_diss_049.png
writing complete_diss_095_Im0.tif
...
writing complete_diss_108_Im1.tif
writing complete_diss_114_Im0.
unexpected panic attack: runtime error: invalid memory address or nil pointer dereference
```

Checking for a nil `io.Reader` is a reasonable change to stop a program from completely crashing. However, I do suspect there is another bug somewhere in the image extraction mechanism.

Included are outputs from my server as well as the output of `pdfcpu validate`:

* [production.txt](https://github.com/j0hax/pdfcpu/files/9242831/production.txt)
* [validate.log](https://github.com/j0hax/pdfcpu/files/9242833/validate.log)

### DIFF
--- a/pkg/pdfcpu/writeImage.go
+++ b/pkg/pdfcpu/writeImage.go
@@ -847,6 +847,9 @@ func RenderImage(xRefTable *XRefTable, sd *StreamDict, thumb bool, resourceName 
 
 // WriteReader consumes r's content by writing it to a file at path.
 func WriteReader(path string, r io.Reader) error {
+	if r == nil {
+		return errors.New("pdfcpu: WriteReader: Please provide r")
+	}
 	w, err := os.Create(path)
 	if err != nil {
 		return err


j0hax:
@hhrutter friendly ping

hhrutter:
Hello!

Your issue is concerning a particular image you are about to extract.
Somehow calling RenderImage in your case returns a nil io.Reader.
The only way to fix this is by reproducing it and thus getting to the root of this.

```
func RenderImage(xRefTable *XRefTable, sd *StreamDict, thumb bool, resourceName string, objNr int) (io.Reader, string, error) {
	// Image compression is the last filter in the pipeline.

	f := sd.FilterPipeline[len(sd.FilterPipeline)-1].Name

	switch f {

	case filter.Flate, filter.CCITTFax, filter.RunLength:
		return renderFlateEncodedImage(xRefTable, sd, thumb, resourceName, objNr)

	case filter.DCT:
		if sd.CSComponents == 4 {
			return renderDCTToPNG(xRefTable, sd, thumb, resourceName, objNr)
		}
		return bytes.NewReader(sd.Content), "jpg", nil

	case filter.JPX:
		return bytes.NewReader(sd.Content), "jpx", nil
	}

	return nil, "", nil
}

func WriteImage(xRefTable *XRefTable, fileName string, sd *StreamDict, thumb bool, objNr int) (string, error) {
	r, _, err := RenderImage(xRefTable, sd, thumb, fileName, objNr)
	if err != nil {
		return "", err
	}
	return fileName, WriteReader(fileName, r)
}
````

hhrutter:
Hi, please get the latest release and check if your issue persists.
If so, then I would need your file for analysis in order to fix this.
Thank you!

hhrutter:
Closed due do inactivity.

PR #518, opened 2022-09-28
Update Dockerfile

mbrevda:
It seems the go package now included a pre-built binary? (Or perhaps that's part of the install process? I don't know how go works...). Regardless, the dockerfile needs to be updated.

As an aside, it might be beneficial to publish a docker image. Aside from the convenience of using the container, it can also be a useful way for systems like [Renovate Bot](https://github.com/renovatebot/renovate) to track releases and keeps deps up to date.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,7 @@
 FROM golang:latest as builder
 
 # install
-RUN go get github.com/pdfcpu/pdfcpu/cmd/...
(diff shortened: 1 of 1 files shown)


mbrevda:
Might need to add `gcompat` for this to work on Alpine (`apk add gcompat`)

hhrutter:
The Alpine based image is not officially supported by the Go project.
Thank you for yor contribution 💚

PR #546, opened 2023-01-04
Ignore null characters

tmdh:
I have a valid pdf file where pdfcpu fails validation because it doesn't handle null characters. Added fix.

### DIFF
--- a/pkg/pdfcpu/string.go
+++ b/pkg/pdfcpu/string.go
@@ -157,8 +157,8 @@ func Unescape(s string, enc bool) ([]byte, error) {
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


hhrutter:
Hello,

There was already work done about the 0x00 char issue.
Have you been using the latest commit?
Can you share a PDF going along for testing?

Thank you!

tmdh:
Yes, I was using the latest commit. Here is a sample pdf:
[nulltest.pdf](https://github.com/pdfcpu/pdfcpu/files/10492704/nulltest.pdf)

hhrutter:
This should be resolved as of v0.4.0

Issue #567, opened 2023-03-02
Unable to go get latest pdfcpu release v0.4.0 to build into golang application


sereeena:
OS is Ubuntu 20.04.5 LTS (focal)
I have upgraded to go version 1.20.1

Was previously using pdfcpu v0.3.13 successfully with go get github.com/pdfcpu/pdfcpu and using the API functions, but I wanted to upgrade to the latest version to use the Resize command. 

When I attempt upgrade I get:
```
go get -u github.com/pdfcpu/pdfcpu
go: github.com/pdfcpu/pdfcpu@v0.4.0: verifying go.mod: github.com/pdfcpu/pdfcpu@v0.4.0/go.mod: reading https://sum.golang.org/lookup/github.com/pdfcpu/pdfcpu@v0.4.0: 404 Not Found
````

Is pdfcpu no longer supported as a library? I don't want to install and use it via command line, it's much better to use the API functions within my golang application, please continue to allow this?

thank you

**Comment by jerem17** [comment]
Same thing here...

**Comment by hhrutter** [comment]
Sorry.. I need to look into this.

**Comment by hhrutter** [comment]
This issue is still under investigation.

workaround for now: `brew install pdfcpu`

**Comment by jerem17** [comment]
Yes but I can't use the API functions directly as if I had installed it with go get.

**Comment by hhrutter** [comment]
👍 my bad, `brew` only installs the cli.

`go get`seems to be running into timeout issues because of the reposize...

Follow the install instrucions in the Readme.md, this seems to be working:
```
git clone https://github.com/pdfcpu/pdfcpu
cd pdfcpu/cmd/pdfcpu
go install
```

jerem17:
Work fine with go get now, thx a lot

hhrutter:
👍  You got lucky... 
Occasional timeouts during `go get` may happen due to current repo size.
For these cases the workaround above still applies until further notice.

sereeena:
go get started working and I was able to get v0.4.0, but the MergeRaw function is not available in the API for this version? The Merge function used to take ReadSeeker inputs and now requires filenames to read from disk.

I think it is supposed to be, as you have written about it in the Release Notes that we should use MergeRaw, but this is not possible?

hhrutter:
You need to get the latest commit.

lpic10:
[...]

PR #574, opened 2023-03-06
Simplify install command

skelouse:


hhrutter:
I am sorry, not  right now..
See #567

PR #594, opened 2023-03-21
fix order of args for CreateFile

### LINKED ISSUE #593: api: create cmd fails
# Issue
Running create causes an error.

```sh
./pdfcpu create /srv/slideshow-exporter/c37d8b52-7fc9-4a89-aa96-afde7c370823/a.json a.pdf
open : no such file or directory
```

#OS
- linux
- mac

# Version
- latest release
- master branch
```
go run ./ version
pdfcpu: v0.4.0 dev
```

I have a PR fix for it

freshteapot:
# What
Running the sub-command "create" works without error

Resolves #593


hhrutter:
Thanks for discovering this!
Unfortunately there was more going an, that's also why the build failed.
See #593

Issue #609, opened 2023-04-19
watermark position is not exact


adamgreenhall:
I was trying to place an image on a PDF at *exactly* a certain offset. After doing some measuring, I found that it was slightly off. This is a problem for my unusual workflow because I am converting the PDF to an image at 600dpi (for later processing and printing).

After digging around, I found that pfdcpu was converting the watermark/stamp offsets to integers internally, at 72dpi (so the max offset error for my use case would be `0.5*72dpi/600dpi*25.4mm/in = 1.524mm`). The [offset math](https://github.com/pdfcpu/pdfcpu/pull/610/files#diff-d0636cb521143ee6cafcf10dcb324edc0f6cb33ef09d9daac8ef13585b31b632L363-L364) is performed in floats - this just appears to be a parsing issue.

PR #610, opened 2023-04-19
compute watermark position in floats, not ints

adamgreenhall:
Fixes https://github.com/pdfcpu/pdfcpu/issues/609

Simply convert the watermark/stamp parser to use floats instead of ints


hhrutter:
👍 
looking good.
I will ensure all offsets and margins as well will get parsed as floats across the board for all commands with the upcoming release.
Thanks for your contribution!

Issue #650, opened 2023-07-12
Infinite loop on pdf with circular bookmark


semvis123:
PDFCPU gets stuck in an infinite loop on a pdf with circular bookmark.
Expected behavior: it detects the loop and throws an error.
Platform: MacOS 14.0
PDF: https://github.com/chromium/pdfium/blob/main/testing/resources/bookmarks_circular.pdf

hhrutter:
Your file is damaged.
It does not open neither in Mac Preview nor in Adobe Reader.

semvis123:
Yes, but pdfcpu should still produce an error instead of endlessly processing it. Current behavior is vulnerable to an DOS attack.

hhrutter:
Have you tried using the latest commit fd12f8b?

I agree that a process hang is bad but what does that have to do with a DOS attack?

semvis123:
I tried https://github.com/pdfcpu/pdfcpu/commit/fd12f8b0f2e3bff59f736070c34d88d84bdb63d7 , same result. When using this library on untrusted user input in a shared application, this bug can cause a problem for other users that are trying to use the service.

semvis123:
https://github.com/chromium/pdfium/blob/main/testing/resources/bookmarks_circular.pdf
or raw:  https://github.com/chromium/pdfium/blob/main/testing/resources/bookmarks_circular.pdf?raw=true

hhrutter:
This is fixed with https://github.com/pdfcpu/pdfcpu/releases/tag/v0.4.2

PR #651, opened 2023-07-12
Fix infinite loop on circular reference

semvis123:
Possible fix for https://github.com/pdfcpu/pdfcpu/issues/650


hhrutter:
This is fixed with https://github.com/pdfcpu/pdfcpu/releases/tag/v0.4.2
Since this release is centered around outline dict handling, fixing this went along with it.

Issue #499, opened 2022-07-22
api: Addwatermark concurrency


Chennnnnnnnnnnnnnnnnn:
An error was reported in 'Addwatermark' with concurrent scenarios："fatal error: concurrent map writes"

position：/pkg/pdfcpu/createText.go:419
ttf.UsedGIDs[gid] = true

maybe use sync.map
or how can i resolve it?

hhrutter:
Hello!

Please explain your usecase in order to reproduce this.

Thank you for using pdfcpu 💚

Chennnnnnnnnnnnnnnnnn:
Thanks for reply

It's a REST API for add PDF watermark, error can reproduce when concurrency more than one

```
if _, ok := font.UserFontMetrics["NSimSun"]; !ok {
        font.LoadUserFonts()
}

wm, err := pdfcpu.ParseTextWatermarkDetails(doc.WaterMarkStr, "sc:2 abs, op:.4, pos:c, points:30, rotation:45, fontname:NSimSun", onTop, 1)
	if err != nil {
		return outfile, err
	}

	inFile := bytes.NewReader(doc.File)
	outFile := bytes.NewBuffer(outfile)

	err = api.AddWatermarks(inFile, outFile, nil, wm, nil)
	if err != nil {
		return outfile, err
	}
```

Error info:
```
fatal error: concurrent map writes

```

hhrutter:
Please post the stacktrace you are getting for further analysis.

hhrutter:
This should be fixed with latest release.

semvis123:
Isn't fully fixed yet. The `UserFontMetrics` is still being used without a lock.
```
fatal error: concurrent map writes

goroutine 29 [running]:
github.com/pdfcpu/pdfcpu/pkg/font.LoadUserFonts()
        /Users/user/go/pkg/mod/github.com/pdfcpu/pdfcpu@v0.4.0/pkg/font/metrics.go:247 +0x24a
github.com/pdfcpu/pdfcpu/pkg/pdfcpu/model.EnsureDefaultConfigAt({0xc005e1c060?, 0xc004f7d900?})
        /Users/user/go/pkg/mod/github.com/pdfcpu/pdfcpu@v0.4.0/pkg/pdfcpu/model/configuration.go:251 +0x150
github.com/pdfcpu/pdfcpu/pkg/pdfcpu/model.NewDefaultConfiguration()```

PR #652, opened 2023-07-12
Fix concurrent write to map userfontmetrics

semvis123:
UserFontMetrics map is not safe to use concurrently, this pull request uses a lock to ensure it is thread safe.
Fixes: https://github.com/pdfcpu/pdfcpu/issues/499

### DIFF
--- a/pkg/api/font.go
+++ b/pkg/api/font.go
@@ -200,7 +200,9 @@ func planeString(i int) string {
 // CreateUserFontDemoFiles creates single page PDF for each Unicode plane covered.
 func CreateUserFontDemoFiles(dir, fn string) error {
 	w, h := 7800, 7800
+	font.UserFontMetricsLock.RLock()
 	ttf, ok := font.UserFontMetrics[fn]
+	font.UserFontMetricsLock.RUnlock()
 	if !ok {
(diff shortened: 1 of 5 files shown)


hhrutter:
Hello!
Do you think you can provide/use a  `func font.GetUserFontMetricsForRead() map[string]TTFLight` wrapping the mutex calls?

semvis123:
Doing so would defeat the purpose of the lock if I'm not mistaken. (unless you return a copy of the map, which is more expensive). A working alternative would be to provide `func font.GetUserFontMetric(name string) TTFLight`. Which would first lock the mutex, then read the map, and finally unlock the mutex and return the `TTFLight`. But we won't be able to use it in the places where it loops through the map, because the whole loop has to have the lock around it. So I don't know if it is worth it in the end.

semvis123:
By using the wrapper you are indeed using the same lock, but you're unlocking the lock before you've actually read data from the map (which defeats the purpose of the lock).

In my use case I run multiple go routines that are all starting pdfcpu jobs, when there are multiple go routines starting jobs they all try to load the metrics cache at once. When the timing is perfect this causes the panic.

hhrutter (approved):

hhrutter:
You were right and I was wrong.

PR #683, opened 2023-08-24
Fix handling of State/StateModel validation

### LINKED ISSUE #682: Error parsing document with State=Cancelled and StateModel=Review
Greetings!  Thank you for all your hard work in maintaining this great library.  I recently bumped into an issue processing a document.  The error message that came back was `pdfcpu: validateStringEntry: dict=Text entry=State invalid dict entry`.  I stepped through with my debugger and found that the value getting rejected was `Cancelled`, which should be a valid state given that the document has a StateModel of `Review`.

I did some more digging and found [this SO answer](https://stackoverflow.com/a/37656612/513769) that details the relationship between `State` and `StateModel`.  I believe that the handling in this library is missing a couple of states and some of the conditional validation of the model.

I'll submit a PR shortly.

## Environment
OS: MacOS Ventura 13.5
Go Version: go1.20.2 darwin/arm64

stevevls:
Fixes #682

This PR extends the validation of `State` and `StateModel` to account for the fact that the acceptable state values depend on the state model.

There is a bit of a chicken-and-the-egg in validating because `StateModel` is required if `State` is set but is optional otherwise.  The approach, then, is to initially accept any valid State and then to do a second check against a specific list of States depending on the StateModel.

My understanding is that the valid combos are as follows in the table, but I could be wrong!

| State Model | States |
| ------------ | ------ | 
| Marked | Marked, Unmarked |
| Review | Accepted, Rejected, Cancelled, Completed, None | 

## Testing
I ran the unit tests...most ran successfully.  The two that failed seemed unrelated to this change.  Nonetheless, I would love suggestions on how to better test this code.

I'm working on crafting a test document.  I'm unable to share the document that caused me to bump against this issue.  My wife has a licensed copy of Adobe Acrobat, so I'm hoping I can borrow her computer to create some examples.


hhrutter:
Thanks for discovering this 👍🏻 
I cleaned up a little and added a required nil check.
Please review.
If you can manage a small test file using this, just put it into pkg/testdata.
The test suite will pick it up.

stevevls:
Code changes look good to me.  It took me a while to generate a test document, but I justed pushed it up.  I ended up spinning up a java project and modifying the example code in the [StackOverflow question](https://stackoverflow.com/questions/37652181/how-do-i-mark-unmark-write-as-accepted-rejected-cancelled-a-text-annotation-usi) I linked in the issue. 😅   I think this feature relates to so document review workflow in old versions of Acrobat, so it's pretty rare.

hhrutter:

PR #686, opened 2023-08-27
Simpler install intructions using `go install` directly

### LINKED ISSUE #685: Readme: Simplify installation from source directly through `go install`
When you install from source you can use `go install` directly instead of going through a manual git clone process.
I prefer this over needing to git clone the whole repo first and is also quicker than getting any binaries.

```
go install github.com/pdfcpu/pdfcpu/cmd/pdfcpu@latest
```

I think this may also be of use for others to quickly copy.
Absolutely minor, I was tempted to just create a PR, but you asked for a issue so here it is

schicho:
Fixes #685

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,11 @@ $ cd pdfcpu/cmd/pdfcpu
 $ go install
 $ pdfcpu version
 ```
+or directly through Go install:
+
+```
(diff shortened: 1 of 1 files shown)


hhrutter:
Absolutely 👍🏻

PR #690, opened 2023-08-28
Don't render table border when width equal zero

### LINKED ISSUE #689: Render borders with width 0
When we create a table specifying border.width = 0, the borders and the grid are still drawn.

I will create a merge request to fix this problem.

roromix:
Fixes #689


hhrutter:
First let's go over the issue in #689

hhrutter:
I wish it was simpler.
See explanation @ #689

PR #694, opened 2023-08-31
support table column padding

roromix:
Fixes #692

PR #697, opened 2023-08-31
PaperSize Change Type

### LINKED ISSUE #695: api create: paper format does not use default orientation
When I create several PDFs in succession with the API, starting with an A4L paper size, and then an A4 size, the Landscape orientation is used instead of the Portrait orientation (default value)

roromix:
Fixes #695


hhrutter:
These are major changes!
Please wait with a PR until the issue is properly discussed.
See also the PR instructions.
Thank you!

hhrutter:
See #694

PR #707, opened 2023-09-07
Fix docs: pdfcpu export → pdfcpu form export

maltsev:
I noticed that there are several places with the command `pdfcpu export` in the docs. But this command doesn't exist. I guess, it should be `pdfcpu form export`.


hhrutter:
Good catch! Thx

Issue #472, opened 2022-05-17
PDF watermark from Reader


vsenko:
Looks like currently it's impossible to create PDF watermarks via Reader (or byte slice, etc.), only from file: https://github.com/pdfcpu/pdfcpu/blob/e89570a4cdffa2129e927ae3aedf1f30cfdc7795/pkg/pdfcpu/stamp.go#L1192

It would be really handy to add such a feature. Furthermore it's already there  for Image watermarks: https://github.com/pdfcpu/pdfcpu/blob/e89570a4cdffa2129e927ae3aedf1f30cfdc7795/pkg/pdfcpu/stamp.go#L1224

hhrutter:
Hello!

I am not quite sure what you are looking for.
Please explain your specific usecase.

In case you are looking for something similar to:
`api.ImageWatermarkForReader(r io.Reader, desc string, onTop, update bool, u pdfcpu.DisplayUnit)`

maybe:
`api.PDFWatermarkForReader(r io.Reader, desc string, onTop, update bool, u pdfcpu.DisplayUnit)`

then that can't be done.

A PDF resource for stamping/watermarking is a combination of some content stream bytes, its resources and a bounding box whereas an Image resource can be repesented by a simple byte slice. 

On top of that pdfcpu supports multi-stamps where each page of the input PDF gets stamped with a different page of the  PDF stamp source file. See more here: https://pdfcpu.io/core/stamp

vsenko:
Excuse me for being that brief, consider the following usecase: the watermark is generated on-the-fly for each incoming PDF from some metadata, thus it's inconvenient to save the watermark beforehand to the file. The code that generates the watermark is on another process/host.

Thank you for identifying the problem with `io.Reader` approach, but maybe we can go with `io.ReadSeeker`:
```
api.PDFWatermarkForReader(r io.ReadSeeker, desc string, onTop, update bool, u pdfcpu.DisplayUnit)
```

vsenko:
Another option (to keep the interface more consistent) is to copy data from `io.Reader` into `io.ReadSeeker` under the hood, but it will require twice more memory during the copy.

hhrutter:
Generating a pdfcpu PDF watermark somewhere in order to use it later on for stamping is a valid use case but can't be done easily. 

[...]

PR #732, opened 2023-11-08
Allow the use of stamps/watermarks loaded into memory

vsenko:
## Use stamps/watermarks loaded into memory

Currently it is possible to load PDF stamps/watermarks only from files. In some cases it would be convenient to use already loaded into memory files (similar was discussed in #472).

This PR proposes to extend the `Watermark` model with additional filed `PDF` which can be used to provide an already loaded PDF stamp/watermark as a `io.ReadSeeker`.

The corresponding code to load it in `createPDFResForWM` is also provided.

### DIFF
--- a/pkg/pdfcpu/model/watermark.go
+++ b/pkg/pdfcpu/model/watermark.go
@@ -62,6 +62,7 @@ type Watermark struct {
 	TextLines         []string            // display multiple lines of text.
 	URL               string              // overlay link annotation for stamps.
 	FileName          string              // display pdf page or png image.
+	PDF               io.ReadSeeker       // reader for PDF watermark.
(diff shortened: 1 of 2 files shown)


hhrutter:
Makes sense.
Let me see how we can fit this into the current API.

vsenko:
I'm using `AddWatermarks` from `api`, the actual code is here: https://github.com/sigex-kz/ddc/blob/e5e0816115a33f55720ca3be08ea3f34fc37d06b/ddc.go#L467

hhrutter:
Thank you for your time and efforts!
This is part of https://github.com/pdfcpu/pdfcpu/releases/tag/v0.6.0

Issue #733, opened 2023-11-08
More control over multistamp/multiwatermark placement (in terms of pages)


vsenko:
Currently multistamp/multiwatermark are applied in a fixed manner: the first page of the recipient PDF gets the first page of the multistamp/multiwatermark (1-1), 2-2, 3-3, and etc.

In certain scenarios it would be convenient to be able to arrange placement of multistamp/multiwatermark differently, for example: 1-x, 2-1, 3-2, etc. Or 1-x, 2-x, 3-1, 4-2, etc. To shift the beginning of multistamp/multiwatermark several pages.

hhrutter:
Not necessarily.
The stamp API provides a selectedPages parameter.
Execute on your cmd line for more on this: `pdfcpu selectedpages`

hhrutter:
This fixed with https://github.com/pdfcpu/pdfcpu/releases/tag/v0.6.0

PR #734, opened 2023-11-08
Shift placement ot multistamp/multiwatermark by several pages

vsenko:
A possible implementation of #733.

The idea is to be able to configure the `Watermark` for how many pages have to be skipped before applying multistamp/multiwatermark.

My understanding of the internals of PDF is not that good, thus I suspect that this implementation is quite brute force.

### DIFF
--- a/pkg/pdfcpu/model/watermark.go
+++ b/pkg/pdfcpu/model/watermark.go
@@ -64,6 +64,7 @@ type Watermark struct {
 	FileName          string              // display pdf page or png image.
 	Image             io.Reader           // reader for image watermark.
 	Page              int                 // the page number of a PDF file. 0 means multistamp/multiwatermark.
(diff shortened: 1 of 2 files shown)


hhrutter:
Hello!

Thanks for the effort.
Generally I expect a conversation before any PR, to understand and see if there is a need or this at all.

Generally the stamp API provides a parameter where you can pass in selectedPages.
Only selected pages will be stamped.

Please describe you use case if this does not suffice your needs.

hhrutter:
Thanks for your time and input!

I have elaborated on your idea and pdfcpu now allows for specifying both the start page # of the stamp file and also the first page number of the destination file to apply multi stamping.

Whenever possible any extension of a pdfcpu command also needs to be reflected on the CLI.
The config syntax for PDF stamp has been extended therefore. See `pdfcpu help stamp`.

Issue #751, opened 2023-12-10
pdfcpu version panics with runtime error


afh:
When trying to update the [pdfcpu package in nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/graphics/pdfcpu/default.nix) to 0.6.0 the installCheck fails. The issue is reproducible when deactivating the installCheck and running `pdfcpu version` manually:

```sh
% pdfcpu version
pdfcpu: v0.6.0 dev
panic: runtime error: slice bounds out of range [:8] with length 1

goroutine 1 [running]:
main.printVersion(0x140001e52b0)
        github.com/pdfcpu/pdfcpu/cmd/pdfcpu/process.go:143 +0x24c
main.commandMap.process(0x140001c5ec8?, {0x16d413117, 0x7}, {0x0, 0x0})
        github.com/pdfcpu/pdfcpu/cmd/pdfcpu/cmd.go:143 +0x2e4
main.main()
        github.com/pdfcpu/pdfcpu/cmd/pdfcpu/main.go:56 +0xac
```

### Thank you for submitting a possible bug!

Please ensure the following:

- [x] Your issue is based on the latest ~~commit~~ release
- [x] State your OS and OS version
```sh
% sw_vers
ProductName:            macOS
ProductVersion:         14.1.2
BuildVersion:           23B92
```

* **N/A** ~~When reporting a problem with a specific PDF input file please avoid stating the organization responsible for the PDFWriter - just refer to the *PDFWriter*~~

hhrutter:
Please pull the release again and let me know if that works for you.
I had to push force the latest commit around noon GMT today.

afh:
Thanks for the info, @hhrutter, much appreciated! When building locally from a clone of this repo `pdfcpu version` works as expected. I'll dig deeper into the issues when building pdfcpu 0.6.0 via nixpkgs and will report back.

afh:
@hhrutter looking at the implementation of [`printVersion`](https://github.com/pdfcpu/pdfcpu/blob/master/cmd/pdfcpu/process.go#L122) it seems the code expects to always be build from a git repo clone and not from a [distribution archive](https://github.com/pdfcpu/pdfcpu/archive/refs/tags/v0.6.0.tar.gz).

[...]

PR #752, opened 2023-12-10
Fix building from distribution archive

afh:
For details please see #751

### DIFF
--- a/cmd/pdfcpu/process.go
+++ b/cmd/pdfcpu/process.go
@@ -132,6 +132,9 @@ func printVersion(conf *model.Configuration) {
 			for _, setting := range info.Settings {
 				if setting.Key == "vcs.revision" {
 					commit = setting.Value
+					if len(commit) >= 8 {
+						commit = commit[:8]
+					}
 				}
 				if setting.Key == "vcs.time" {
 					date = setting.Value
@@ -140,7 +143,7 @@ func printVersion(conf *model.Configuration) {
 		}
 	}
 
-	fmt.Fprintf(os.Stdout, "commit: %s (%s)\n", commit[:8], date)
+	fmt.Fprintf(os.Stdout, "commit: %s (%s)\n", commit, date)
 	fmt.Fprintf(os.Stdout, "base  : %s\n", runtime.Version())
 	fmt.Fprintf(os.Stdout, "config: %s\n", conf.Path)
 }


afh:
@hhrutter I thought a bit more about this PR and think that the latest proposed changes here are much less disruptive to pdfcpu's `version` output, while avoiding a panic, if `commit` does not contain expected data, thus allowing distributions to provide commit info from the outside instead of a git repo.

hhrutter:
😞 This way pdfcpu loses control about the `version` output.
My stance is, if debug.ReadInfo is not available then the build has failed - this approach does not serve this intention.

The output of `pdfcpu.version` for a particular version is set in stone and must be identical no matter the platform.

The options are you build straight from the repo or grab the binary from Github.
You do this once for every release, so maybe 4 or 5 times a year..
I wonder how other Linux distros deal with this.

afh:
I see your point and respect your stance, @hhrutter, and am trying to find the best compromise between pdfcpu's and nixpkgs (potentially somewhat contradicting) preferences and requirements.

Note that I'm an interested user and keen learner of Nixpkgs and NixOS, but by no means (yet?) deeply involved in the project, so my knowledge is limited and from that my assumptions could be wrong. If my statements are incorrect, I'm happy to be corrected 🙂 

Nixpkgs is a source based distribution and strives for [reproducible builds](https://reproducible.nixos.org), so grabbing the binary from GitHub does not serve that purpose.
While pre-build binaries are available for NixOS and Nixpkgs from my experience Nix builds software more often than not (especially with [Nix Flakes](https://nixos.wiki/wiki/Flakes)), so pdfcpu's Git repository would likely be downloaded from many, many more users than just the Nix organisation and much, much more frequently than 4 or 5 times a year.

I've proposed a few options (see [my comment on NixOS/nixpkgs#273438](https://github.com/NixOS/nixpkgs/pull/273438#discussion_r1424981192)), let's see where this leads to.

afh:
I'll keep an eye out for pdfcpu 0.6.0 on other distributions, if you know of any that have already updated, please let me know.

afh:
Thank you for providing additional details on Nix, @doronbehar, very much appreciated. If I wanted to read more on the topic where should I start?

@hhrutter I'll leave this PR open as I believe it improves the code by avoiding a panic, even at the cost that the commit may not show the intended/desired information.
Please merge or discard it as you see fit 🙂 

As for the pdfcpu update to 0.6.0 in Nixpkgs, it is build from the git source and the `version` command will show the commit and date values from the git clone.

Issue #761, opened 2023-12-24
bugs found by the linter


sivukhin:
I ran [govanish](https://github.com/sivukhin/govanish) linter (it still in the WIP phase) against `pdfcpu` repo and it found several suspicious places. I don't use `pdfcpu` and don't have much context about the project - but found issues looks like a bugs in the implementation.

Linter found following issues:
1. `StreamDict.Clone` method has following code snippet
```go
f := PDFFilter{}
f.Name = v.Name
if f.DecodeParms != nil {
	f.DecodeParms = v.DecodeParms.Clone().(Dict)
}
pl[k] = f
```
  `govanish` reported that `if` condition were removed by compiler which is correct behaviour - `f.DecodeParms` is always `nil` (we created it in the loop). It seems like this is a typo and `v.DecodeParms` should be used there
2. `ParseXRefStreamDict` method has following code snippet:
```go
if len(indArr)%2 > 1 {
	return nil, errXrefStreamCorruptIndex
}
```
  `govanish` reported that this `if` condition were eliminated by compiler and this is correct - because `x % 2` always greater than 1. Based on the code after that check I suspect that this check supposed to validate that array has even length.
3. In the `equalFontNames` internal helper there is a forgotten type cast check.

hhrutter:
Thanks I'll take a look at this.

PR #762, opened 2023-12-24
Fix bugs

sivukhin:
PR for the Issue #761 

Changes: 
1. replace `f.DecodeParms != nil -> v.DecodeParms != nil` in the `StreamDict.Clone` method
2. replace `len(indArr)%2 > 1 -> len(indArr)%2 != 0` in the `ParseXRefStreamDict` method
3. add missed type check in the `equalFontNames`

### DIFF
--- a/pkg/pdfcpu/model/equal.go
+++ b/pkg/pdfcpu/model/equal.go
@@ -21,8 +21,9 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/pdfcpu/pdfcpu/pkg/pdfcpu/types"
 	"github.com/pkg/errors"
+
+	"github.com/pdfcpu/pdfcpu/pkg/pdfcpu/types"
 )
 
 // EqualObjects returns true if two objects are equal in the context of given xrefTable.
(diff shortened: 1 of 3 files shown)


joel-rieke:
Interestingly enough, adding this code:
	name, ok := o.(types.Name)
	if !ok {
		nameInt, ok := o.(types.Integer)
		if !ok {
			return nil, errors.Errorf("pdfcpu: validateNameEntry: dict=%s entry=%s invalid type %T", dictName, entryName, o)
		}
		name = types.Name(nameInt.String())
	}

Seems to make the issue https://github.com/pdfcpu/pdfcpu/issues/780 work.... No idea if that's even valid... but it seems to make the pdf validate..?

pkg/pdfcpu/model/equal.go
```diff
@@ -141,7 +142,7 @@ func equalFontNames(v1, v2 types.Object, xRefTable *XRefTable) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	bf2 := v2.(types.Name)
+	bf2, ok := v2.(types.Name)
 	if !ok {
 		return false, errors.Errorf("equalFontNames: type cast problem")
```
joel-rieke, line 147: If this is added, it fixes parsing for another pdf...
name, ok := o.(types.Name)
if !ok {
nameInt, ok := o.(types.Integer)
if !ok {
return nil, errors.Errorf("pdfcpu: validateNameEntry: dict=%s entry=%s invalid type %T", dictName, entryName, o)
}
name = types.Name(nameInt.String())
}

hhrutter:
Thank you for your contribution! 🎉

PR #776, opened 2024-01-15
Avoid calling "DecodeName" when parsing dictionaries.

### LINKED ISSUE #775: Parsing file with lots of dictionaries is extremely slow
I experienced this with a PDF that is a 9-page export of CAD drawings containing lots of small lines and numbers (created using "PDFTron PDFNet, V6.40292"). Unfortunately I can't share the document.

Tested with latest master (b89d7b1ab108d54a983c9653d47df07bbad5336e) and the test script below.

<details>

```golang
package main

import (
	"log"
	"os"

	"github.com/pdfcpu/pdfcpu/pkg/pdfcpu"
	"github.com/pdfcpu/pdfcpu/pkg/pdfcpu/model"
)

func main() {
	log.SetFlags(log.Flags() | log.Lmicroseconds)
	fp, err := os.Open("slow-parse.pdf")
	if err != nil {
		log.Fatal(err)
	}
	defer fp.Close()

	conf := model.NewDefaultConfiguration()
	log.Printf("Parsing ...")
	pdf, err := pdfcpu.Read(fp, conf)
	if err != nil {
		log.Fatal(err)
	}
	log.Printf("Done")

	if err := pdf.EnsurePageCount(); err != nil {
		log.Fatal(err)
	}

	log.Printf("Parsed %d pages", pdf.PageCount)
}
```

</details>

Parsing doesn't stop (I killed it after 5 minutes taking 100% CPU). I have a patch ready takes improves parsing this file to below 4 seconds.

fancycode:
If no name with a "#" has been added before, it's not necessary to go through the expensive "Insert" call that will call "DecodeName" for every previously added key.

Fixes #775

Even better would be to keep the `hasNames` flag in the `Dict` object so other calls to `Find` could also benefit, but this is not possible with the aliased type.

### DIFF
--- a/pkg/pdfcpu/model/parse.go
+++ b/pkg/pdfcpu/model/parse.go
@@ -530,6 +530,7 @@ func parseName(line *string) (*types.Name, error) {
 func processDictKeys(line *string, relaxed bool) (types.Dict, error) {
 	l := *line
 	var eol bool
+	var hasNames bool
 	d := types.NewDict()
 	for !strings.HasPrefix(l, ">>") {
 		key, err := parseName(&l)
@@ -558,8 +559,18 @@ func processDictKeys(line *string, relaxed bool) (types.Dict, error) {
 			if log.ParseEnabled() {
 				log.Parse.Printf("ParseDict: dict[%s]=%v\n", key, obj)
 			}
-			if ok := d.Insert(string(*key), obj); !ok {
-				return nil, errDictionaryDuplicateKey
+			stringKey := string(*key)
+			if !hasNames {
+				// Avoid expensive "DecodeName" on existing keys in "Insert".
+				if _, found := d[stringKey]; found {
+					return nil, errDictionaryDuplicateKey
+				}
+				d[stringKey] = obj
+				hasNames = strings.IndexByte(stringKey, '#') >= 0
+			} else {
+				if ok := d.Insert(stringKey, obj); !ok {
+					return nil, errDictionaryDuplicateKey
+				}
 			}
 			continue
 		}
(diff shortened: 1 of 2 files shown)


fancycode:
Timings from my test program:
```
$ go run test.go 
2024/01/15 10:34:07.566998 Parsing ...
2024/01/15 10:34:10.899487 Done
2024/01/15 10:34:10.899501 Parsed 9 pages
```

hhrutter:
Thanks for putting in the work, although I'd rather discuss issue resolutions before pull requests get filed.
I am evaluating this.

fancycode:
> Thanks for putting in the work, although I'd rather discuss issue resolutions before pull requests get filed.

Well, the fix was pretty straightforward so I didn't expect much discussion. But I'm happy if you come up with a better solution.

joel-rieke:
If this one works, wow, that's a lot faster.  What are the side effects on optimization if any?

fancycode:
Just updated the test to better match the contents of my problematic file. In the file there are two dictionaries with about 200.000 entries each. The test generates a dictionary with 50.000 entries and now takes (on my machine) ~29 seconds on master and ~0.05 seconds with this PR.

While this PR fixes this particular problem, you could still construct a file that takes very long to parse.

hhrutter:
Excellent patch!
Thanks for your contribution 💚

fancycode:
Thanks for pdfcpu, glad to be able to give something back!

hhrutter:
I believe we need to walk back some of these changes:

parse



First of all in line 594 we need to check the return code.

This is all about recognizing existing dict keys and also processing any embedded 2 digit hexcodes in Name objects like in:
`/A#42` which (bytewise) is the same as `/AB`.

Parsing both << /A#42 (A) /AB (B) >> and << /AB (B) /A#42 (A) />> should produce a Duplicate Key error which is not the case.

There is also a bug in the existing dict.Find(key) method where the key should also be decoded before comparing
since it's a name. That's on me :)

I have to take a step back and think about how to best fit in these pieces - processDictKeys is already way too complex.
I need to think more about this but this defnitely can't stay like this.
So consider this a headsup.

PS: I still think we can make some shortcuts here like per your proposal but it's gotta be different.

hhrutter:
Correcting myself.

If there are dict keys using hex codes, then we really only need to support locating them by the original key or a normalized version that contains 2 bytes for each # sequence.

So as per your idea we should be fine! 👍🏻 

Also, I checked my local test corpus and there is only a very small number of files that actually using # within names.
So the excellent news is, parsing gets a major performance boost.  🚀

Still, I refactored the code around that location plus I fixed the case so that DecodeName gets also called on the very first occurence of # in a dict key. 

Hopefully all in your spirit.

So thanks again 🙏🏻 

Please get the latest commit.

Issue #783, opened 2024-01-23
additional booklet options


adamgreenhall:
it turns out there are several more options for booklet making workflows than we originally included in pdfcpu. In particular there are three main types:
* *"simple"* booklet
* *"advanced"* booklet - favored by some book binders that prefer a more complex, but more efficient collation process. This is the type of booklet ordering currently in pdfcpu
* perfect bound booklet - a booklet with a glued binding (i.e. cut the sheets into 4 pages, collate, and then glue), instead of folded and then stapled, like the other two methods above

Each type of booklet comes with its own preferred ordering of pages. 

Also, support for 6-up and 8-up booklets.

I've been using these different methods on a branch for awhile - PR to follow to merge them in.

hhrutter:
That sounds excellent!
👍🏻 Keep em coming..

piotr-lasota:
I might add to that (although perhaps this is me being unable to understand the docs) some additional config for how do we want to print/duplex the output.

e.g. you have a horizontal A5 book that you want to print on A3 horizontal as a booklet folded over the short edge so that you effectively get 8 pages per sheet. Seems like something that can't be done currently.

petervwyatt:
I'd suggest looking at JDF/XJDF specifications if you want to understand the full gamut (and correct terminology) of booklet printing, product folds, page sequencing, etc. See https://github.com/cip4/JDF-Specification - the illustrations (e.g. Table 7.48) are very informative to avoid needing to read the entire spec...

Note that JDF is also directly supported in PDF/VT via DParts/DPM, and "ISO 21812-1:2019 Graphic technology — Print product metadata for PDF files —  Part 1: Architecture and core requirements for metadata".

hhrutter:
#784 merged in.

PR #784, opened 2024-01-23
additional booklet options

adamgreenhall:
https://github.com/pdfcpu/pdfcpu/issues/783

happy to change things around in the code, and/or provide more examples/documentation. Some description of the different booklet types [here](https://spectrolite.app/how-to/zines-and-books/layouts-zines/imposition-types)

also, two unrelated issues - sorry about that - I can pull these out, if you prefer.


pkg/pdfcpu/types/layout.go
```diff
@@ -404,7 +404,8 @@ func ParsePageFormat(v string) (*Dim, string, error) {
 
 	dim := Dim{d.Width, d.Height}
 	if (d.Portrait() && !portrait) || (d.Landscape() && portrait) {
-		dim.Width, dim.Height = dim.Height, dim.Width
```
adamgreenhall, line 407: I found this was causing an error because `Width` and `Height` were getting swapped on `PaperSize[v]`. I forget what the error was though

hhrutter:
Great work 🎉  

Allright, so now we have booklet types, bindings..
I am guessing you use pdfcpu in your Go backend..
Have you thought about extending the booklet command?

If so, do you think you could update the doc page (booklet.md) which is located in the `gh-pages` branch ? 
This would also help me better understand the new booklet features 😉 

Now do we have tests for each of the booklet_types in booklet_test.go ?
Little bit overwhelmed to be honest...

adamgreenhall:
Yes - I'll update the doc page and also the command too. 

This stuff always confuses me too. When developing, I often resorted to making a numbered pages pdf (using stamp) and then double checking things by running it through booklet, printing it out, folding it, and making sure that it was correct. But now there are test cases for everything too.

hhrutter:
Very much appreciated - take all the time you need.

adamgreenhall:
ok CLI options are in and here's a [PR for the website documentation](https://github.com/pdfcpu/pdfcpu/pull/802). 

a few things that came up: 
- the default page ordering has changed for 4-up. Previously for 4-up we were using what is now called `btype=advanced`. But most printers find this advanced method more difficult (and it doesn't apply for the other n), so I think it makes sense to switch the default. 
- as a result of ^, the 4-up example ordering has changed. I corrected the first 4-up pdf example in the docs, but you may want to re-generate it using the original pdf input (with diagonal lines). Also the image example is 4-up and the output image would need to change.
- because of the addition of `binding=short`, the cut and fold guidelines will not be correct for all page+binding orientations. Would you like to keep them? If so I can take a stab at correcting them.

hhrutter:
Thanks Adam, I will check it out!

pkg/pdfcpu/types/layout.go
hhrutter: This is kind of dangerous because it is used by other commands.
I will have to check this.

hhrutter:
In general I think we should keep and update the cut and fold guidelines for all possible booklet variation for consistency.
If that's not possible or too much work, we should document this in the usage & the doc page.

I am quite busy with the upcoming PDF 2.0 support but I'd like the enhanced booklet functionality to be part of the upcoming release. I'd like to leave the touching up part of this to you if that's ok..

Then for new source files even for tests we'd need the license header at the top.

And we still need to figure out the dim swapping issue in layout.go

adamgreenhall:
Sounds good - I will take on adding cut/fold guidelines for the new types.
Also will add missing license headers.

pkg/pdfcpu/types/layout.go
hhrutter: So the original code should be fine for cases where you want to enforce portrait or landscape mode by appending P or L respectively to the papersize name.

adamgreenhall: ah - cool. that makes sense. I think I had a broken version of this function in my own code (without the dim copy) and mistakenly edited this one too. Will remove

adamgreenhall:
@hhrutter ok, I think I have all the things we talked about so far done:

[x] copyright headers
[x] remove `ParsePageFormat` change
[x] booklet guidelines for diff orientations and types

Want to take another look?

hhrutter:
Thanks, yes let me take a look please.

hhrutter:
This was merged in manually.
I hope I did not miss anything - Please double check!
I put in minor modifications mostly around the parsing of the `n` value for the CLI.
Thank you so much for your time working on this  🎉

adamgreenhall:
Doubled checked - looks good. Thanks!

PR #785, opened 2024-01-23
Add new reading / parsing functions that take a Context object.

### LINKED ISSUE #777: Allow cancelling of parsing
Depending on the input file, parsing might take too long (see #775). It would be great if there was a way to cancel a previous call to `pdfcpu.Read`.

One possibility would be to add a `context.Context` to the `m̀odel.Configuration`, that, when cancelled, will stop further parsing and return the error from the context as result of the `pdfcpu.Read` call.

I'm happy to provide a PR for this.

fancycode:
This will allow the caller to optionally cancel long-running parsing calls.

Most of the (internal) changes are just passing the `context.Context` along. In some places, it is checked for an error and if so (e.g. because the context was cancelled for whatever reason), parsing / reading will be interrupted.

The API is backwards compatible, i.e. the current functions still exist and will use a `context.Background()` internally.

Fixes #777

### DIFF
--- a/pkg/pdfcpu/read.go
+++ b/pkg/pdfcpu/read.go
@@ -19,6 +19,7 @@ package pdfcpu
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"io"
 	"os"
 	"sort"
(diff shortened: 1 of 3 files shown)


pkg/pdfcpu/read.go
hhrutter, line 1: What is the effect for callers of Read which is now wrapping s  func using a brand new context.Background() ?
Just for backwards compatibility or any further usage?

fancycode: This is just for backwards compatibility: `context.Background` is never cancelled, so the behaviour of the existing functions will be the same as it currently is.

hhrutter:
Again, thanks! 🎉

PR #795, opened 2024-01-29
Further improve parsing of dictionaries / names

fancycode:
With this change, the names are decoded internally, so they can be compared directly when adding entries to dictionaries. On writing, the names are encoded if necessary.

Also removed some duplicate code for name encoding / decoding and simplified object type tests.

Follow-up to #776 to also speed up parsing dictionaries that contain key with a `#`.


hhrutter:
Is there any way to get into a discussion of an issue.
This especially applies for PRs with heavy, critical changes.

fancycode:
As written above, it's a follow-up to the previous change in #776, related to issue #775.

Even with the change from #776 you can construct a PDF with dictionaries that take ages to parse (see the updated test in https://github.com/pdfcpu/pdfcpu/blob/4978e9c5d74a9a7dc5679fa0a644d4f414797eff/pkg/pdfcpu/model/parse_dict_test.go#L153).

I'm happy to discuss the changes in this merge request - at least that's my understanding what merge requests are for. You can easily comment on individual lines, add a review or add global comments as we are doing right now.

hhrutter:
Do you think you can rebase this onto the latest commit?
Will help big time cutting a new release by the end of the day 😉

fancycode:
Sure, just rebased the branch on fc87a229d25c15f37228377854c7ff56cff153d7.

hhrutter:
heads up... ValidationNone is gone in case you were using it..

hhrutter:
Thanks!

PR #811, opened 2024-02-28
Avoid copying from "bytes.Buffer" to get underlying bytes.

### LINKED ISSUE #810: Avoid unnecessary copying with filters.
The filters usually return a `*bytes.Buffer`, so copying it into a temporary `bytes.Buffer` to get the underlying `[]byte` can be avoided.

Example:
https://github.com/pdfcpu/pdfcpu/blob/fc87a229d25c15f37228377854c7ff56cff153d7/pkg/pdfcpu/types/streamdict.go#L292-L297

Same applies for the filters itself which can be chained and in such casees also don't need to copy from the passed `io.Reader` if it is actually a `*bytes.Buffer`.

fancycode:
As described in #810, this PR removes the unnecessary copying from `bytes.Buffer` to get the `[]byte` data where possible.

Fixes #810

### DIFF
--- a/pkg/filter/ascii85Decode.go
+++ b/pkg/filter/ascii85Decode.go
@@ -33,14 +33,11 @@ const eodASCII85 = "~>"
 // Encode implements encoding for an ASCII85Decode filter.
 func (f ascii85Decode) Encode(r io.Reader) (io.Reader, error) {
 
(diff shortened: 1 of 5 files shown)


hhrutter:
Thanks again! Good catch 👍🏻

PR #873, opened 2024-05-21
Feature:create pdf from PDF structure instead of json files

carmel:
#874


carmel (commented):
Create from PDF struct instead of json

hhrutter:
Thank you for your contribution but a general API call for PDF creation is premature at this point.
See also my comment at #874.

PR #881, opened 2024-05-29
Allow property values containing an equal sign

### LINKED ISSUE #880: Allow properties with an equals sign in value
When investigating possible solutions for #424, I've found out that currently it's not possible to set property values containing an equal sign due to the used Split function. PR follows😊

xelan:
Fixes #880


xelan:
Should I also add an additional test case in `property_test.go`?

hhrutter:
No, that's fine..

xelan:
Thanks for the merge :+1:

PR #895, opened 2024-06-21
fix(stamp): Add Support for Soft Mask in RGBA Image Buffer

### LINKED ISSUE #894: Bug with black background when RGBA image is stamped on PDF
## description

When an image with a transparent background is added to a PDF, such as favicon.png below, the background turns black. (In the case of RGBA)

favicon

スクリーンショット 2024-06-21 11 30 35

## example code
```
func StampImage(inFile string, outFile string, selectedPages []string, fileName string) error {
	config := model.NewDefaultConfiguration()
	desc := "pos: tl, off: 0 0, scale: 1 a, rot:0, mode:1"
	onTop := true
	wm, err := pdfcpu.ParseImageWatermarkDetails(fileName, desc, onTop, types.POINTS)
	if err != nil {
		return err
	}
	return api.AddWatermarksFile(inFile, outFile, selectedPages, wm, config)
}
```

## environment
macOS Ventura 13.0
pdfcpu https://github.com/pdfcpu/pdfcpu/commit/576f15e6cd191b62168d7b26879ab3b1133b29bc

toshi1127:
Fixes #894

## description

When an image with a transparent background is added to a PDF, such as favicon.png below, the background turns black. (In the case of RGBA)

favicon

### before

スクリーンショット 2024-06-21 11 30 35

### after

スクリーンショット 2024-06-21 11 37 36

### DIFF
--- a/pkg/pdfcpu/model/image.go
+++ b/pkg/pdfcpu/model/image.go
@@ -184,15 +184,29 @@ func CreateDCTImageObject(xRefTable *XRefTable, buf []byte, w, h, bpc int, cs st
 	return sd, nil
 }
 
-func writeRGBAImageBuf(img image.Image) []byte {
+func writeRGBAImageBuf(img image.Image) ([]byte, []byte) {
 	w := img.Bounds().Dx()
 	h := img.Bounds().Dy()
 	i := 0
+	var sm []byte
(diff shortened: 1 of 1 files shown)


hhrutter:

Issue #946, opened 2024-09-09
Dockerfile WORKDIR declared as /data not /app


carlwilson:
The Dockerfile currently declares it's final [`WORKDIR` as `/data`](https://github.com/pdfcpu/pdfcpu/blob/master/Dockerfile#L42). This means the `/app` directory documented in the README has to be declared as an absolute path when passing PDFs for validation.

```shell
$ docker run -it --mount type=bind,source="$(pwd)",target=/app pdfcpu pdfcpu validate /app/pdfs/a.pdf
```

The `/app` dir should probably be declared as a container `VOLUME` as well.

hhrutter:
no objection.

carlwilson:
I'll tease the larger PR apart and add one that simply fixes this issue for now.

hamdan7m:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

PR #947, opened 2024-09-09
Use Docker ENTRYPOINT rather than CMD

### LINKED ISSUE #945: Use Docker ENTRYPOINT rather than CMD
Replace use of `CMD` with `ENTRYPOINT` in Dockerfile.

See this [Docker blog](https://www.docker.com/blog/docker-best-practices-choosing-between-run-cmd-and-entrypoint/) for rationale assuming this container is intended as an exclusively `pdfcpu` container.

carlwilson:
- replaced used of `CMD` with `ENTRYPOINT` in Dockerfile;
- fixed the instructions for using the Docker image for `ENTRYPOINT` invocation; and
- expose the `/app` folder as `VOLUME`.

The issues below detail the thinking:

Closes #945, #946


hhrutter:
Hi!
I am not a docker expert/captain..
Is there any way to modify th docker file so the container executes and outputs `pdfcpu version -v`  on startup?

carlwilson:
Hi,
I understand the question and why you're asking it. I fear the answer is no, but I'm happy to try some ideas and get back to you. It might be a couple of weeks as time is currently a little tight,

hhrutter:

PR #960, opened 2024-10-02
Support IndirectRef entries for Checkbox appearance dict (and its values)

rluba:
This fixes filling out checkboxes in forms that use non-standard "Yes" states and define their appearance via IndirectRef entries (as seen in some EPO forms).

Here’s an example of a PDF form that requires this fix to render checked checkboxes correctly: https://link.epo.org/web/applying/forms/epo-form-7000-07-24-editable.pdf
(It’s a request form by the European Patent Office for a Unitary Patent.)

Here’s how to repro the issue:
1. `pdfcpu form export epo-form-7000-07-24-editable.pdf`
2. Edit `out.json` and set one or more of the checkbox values to true.
3. `pdfcpu form fill epo-form-7000-07-24-editable.pdf out.json out.pdf`
4. Look at the result. The checkboxes are not rendered as checked.

The reason is that all but the first checkbox store their `AP` dicts as `IndirectRef` (and even the values inside that dict are `IndirectRef` entries), so the old code fails to detect that the `Yes` state is called `Ja` and therefore uses the wrong value when setting the appearance state (`AS`).

### DIFF
--- a/pkg/pdfcpu/form/fill.go
+++ b/pkg/pdfcpu/form/fill.go
@@ -569,7 +569,7 @@ func fillCheckBoxKid(ctx *model.Context, kids types.Array, off bool) (*types.Nam
 		return nil, errors.New("pdfcpu: corrupt AP field: missing entry N")
 	}
 
-	offName, yesName := primitives.CalcCheckBoxASNames(d2)
+	offName, yesName := primitives.CalcCheckBoxASNames(ctx, d2)
 	asName := yesName
 	if off {
 		asName = offName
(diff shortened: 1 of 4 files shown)


hhrutter:
Thanks, I will check it out.
Next time please do me a favor and create an issue beforehand.
I prefer PRs that fix issues.
Thank you!

hhrutter:
Sorry, I am unable to reproduce your problem using the linked PDF and the current commit.
Filling all checkboxes works as intended.

Issue #968, opened 2024-10-15
booklet 8up orientations


adamgreenhall:
Currently booklet making only supports 8up for portrait orientation with long-edge binding. But there are 3 more options! PRs to follow.

hhrutter:
👍🏻 perfect timing. I am preparing the next release.

hamdan7m:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

PR #969, opened 2024-10-15
Booklet 8up orientations

adamgreenhall:
adds the other 8up orientations
https://github.com/pdfcpu/pdfcpu/issues/968

see also: website changes https://github.com/pdfcpu/pdfcpu/pull/970


adamgreenhall:
@hhrutter - re-testing with the latest change now. The new 8ups look good.

I did notice a problem that (I think) is not related to this change, but happened with the addition of `nup.Enforce`. The basic issue is that for Nup cases where the pages are *intended* to be oriented differently on the sheet from their original. This includes 2up, and also 8up (which is why I noticed it). See an example of this problem in [BookletFromPDFA4_2Up.pdf](https://github.com/pdfcpu/pdfcpu/blob/9b3d855965979cde0edc6d4ed4092f0ec200f457/pkg/samples/booklet/BookletFromPDFA4_2Up.pdf) - it's now preserving the portrait orientation, when it should be oriented landscape on the page (page is still portrait).

You can make it work in these cases by [setting](https://github.com/pdfcpu/pdfcpu/pull/969/files#diff-8016f1ad4f7ef039952152d2f2856982bfa43a3dcddee7a98da1e717b2781099R137) `enforce:true` in the config (see [BookletFromPDFLetter_2Up.pdf](https://github.com/pdfcpu/pdfcpu/blob/9b3d855965979cde0edc6d4ed4092f0ec200f457/pkg/samples/booklet/BookletFromPDFLetter_2Up.pdf)). But this seems like a non-ideal hack. Ideas?

hhrutter:
You are right!
This issue affects all of the following:
```BookletFromPDFA4_2Up.pdf
BookletFromPDFLedger_8Up.pdf
BookletFromPDFLetter_2Up_perfectbound.pdf
BookletFromPDFLetter_2Up.pdf
BookletFromPDFLetter_2UpWithTrailingBlankPages.pdf
HardbackBookFromPDF.pdf
```

Looking into it..

hhrutter:
Ah, looks like the proper enforce configuration for Booklets slipped thru when I commited this.

`enforce` is ok, because it enforces a best-fit approach for nupping.
I am aware the naming is mediocre but it does the job.
It's also used by other commands.

Do you want to include this change in your PR?

```
// DefaultBookletConfig returns the default configuration for a booklet
func DefaultBookletConfig() *model.NUp {
	nup := model.DefaultNUpConfig()
	nup.Margin = 0
	nup.Border = false
	nup.BookletGuides = false
	nup.MultiFolio = false
	nup.FolioSize = 8
	nup.BookletType = model.Booklet
	nup.BookletBinding = model.LongEdge
	nup.Enforce = true
	return nup
}
```

adamgreenhall:
That makes sense, thanks. Everything looks good now - should be good to go.

hhrutter:
Nice addition.
Thank you so much!

PR #999, opened 2024-11-20
Buffer parse partial

### LINKED ISSUE #997: Slow parsing for objects with huge entries in dictionaries

Please ensure the following:

Yes (c4f314bcfb19ecde85b91e846a84ac0c3b9fcc80)

Ubuntu 22.04

We saw a file on one of our servers that contained multiple objects consisting of a single dictionary that had an entry with a value of about 10 MBytes of hex-encoded binary data, i.e.
```
123 0 obj
<<
  /Contents <very-long-hex-encoded-data>
  ...other-entries...
>>
endobj
```

Unfortunately I can't share this file. Parsing this file takes ages (cancelled after a couple of minutes), most likely due to `buffer` reading data incrementally and then searches for `stream` / `endobj` in the buffer:
https://github.com/pdfcpu/pdfcpu/blob/c4f314bcfb19ecde85b91e846a84ac0c3b9fcc80/pkg/pdfcpu/read.go#L1708

I'm working on a fix but already adding as issue for discussion.

fancycode:
Fixes #997. See the tests for an example of objects causing the problems.

You can run the tests without the PR to see the slow parsing.


fancycode:
My previous approach (parsing only the newly read buffers) was not working for various reasons. The new one (increasing the buffer size that gets read in subsequent iterations) works and fixes the issue.

So this is ready for review now.

PR #1007, opened 2024-11-25
Prevent division by zero in extractXRefTableEntriesFromXRefStream

### LINKED ISSUE #1006: Division by zero in extractXRefTableEntriesFromXRefStream

Please ensure the following:

Yes

Ubuntu Linux 20.04

PDF metadata reports Producer: "Corel PDF Engine Version 17.6.0.1021" and "Creator" CorelDRAW X7.

The code to parse the xref streams has a division by zero issue if all entries in the `W` array are zero, leading to `xrefEntryLen` being zero here:
https://github.com/pdfcpu/pdfcpu/blob/b9b3d0fca12d9570a60ef7dd3527971bc13b8f71/pkg/pdfcpu/read.go#L538-L545

I suggest changing the condition to
```go
if xrefEntryLen != 0 && len(buf)%xrefEntryLen > 0 {
```

Or you could omit calling this function completely if the buffer is empty. The file that has all values in the `W` array set to zero calls this function with a `buf` length of zero.

fancycode:
Fixes #1006

### DIFF
--- a/pkg/pdfcpu/read.go
+++ b/pkg/pdfcpu/read.go
@@ -540,7 +540,7 @@ func extractXRefTableEntriesFromXRefStream(buf []byte, offExtra int64, xsd *type
 		log.Read.Printf("extractXRefTableEntriesFromXRefStream: begin xrefEntryLen = %d\n", xrefEntryLen)
 	}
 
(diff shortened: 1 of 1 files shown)


hhrutter:
Great catch!

PR #1010, opened 2024-11-25
Check if "entry.Offset" is nil before dereferencing.

### LINKED ISSUE #1009: nil pointer dereference in processXRefStream

Please ensure the following:

Yes

Ubuntu Linux 20.04

The field `entry.Offset` can be `nil` for compressed entries here, leading to a nil pointer dereference:
https://github.com/pdfcpu/pdfcpu/blob/b9b3d0fca12d9570a60ef7dd3527971bc13b8f71/pkg/pdfcpu/read.go#L647

The `entry` is created in `createXRefTableEntryFromXRefStream` here (note that the field `Offset` is not set as with uncompressed entries above):
https://github.com/pdfcpu/pdfcpu/blob/b9b3d0fca12d9570a60ef7dd3527971bc13b8f71/pkg/pdfcpu/read.go#L510-L515

The solution would be to also check for `entry.Offset != nil` before rereferencing.

fancycode:
Fixes #1009

### DIFF
--- a/pkg/pdfcpu/read.go
+++ b/pkg/pdfcpu/read.go
@@ -644,7 +644,7 @@ func processXRefStream(ctx *model.Context, xsd *types.XRefStreamDict, objNr, gen
 
 	*offset += offExtra
 
-	if entry, ok := ctx.Table[*objNr]; ok && *entry.Offset == *offset {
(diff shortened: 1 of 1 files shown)


hhrutter:
Thanks for the contribution 💚

PR #1013, opened 2024-11-25
Correctly handle empty type field length in cross reference streams.

### LINKED ISSUE #1012: Cross reference streams with empty type field length are not handled correctly

Please ensure the following:

Yes

Ubuntu Linux 20.04.

Producer metadata reports "Microsoft® Word für Microsoft 365", document is signed by SwissID.

The `W` array of cross reference streams indicates the lengths of the different fields.

According to the PDF spec (7.5.8.2 Cross-Reference Stream Dictionary): `If the first element is zero, the type field shall not be present, and shall default to type 1.`

This is wrtten in a comment but not handled in https://github.com/pdfcpu/pdfcpu/blob/b9b3d0fca12d9570a60ef7dd3527971bc13b8f71/pkg/pdfcpu/read.go#L532

The first byte `buf[i]` is always treated as the type:
https://github.com/pdfcpu/pdfcpu/blob/b9b3d0fca12d9570a60ef7dd3527971bc13b8f71/pkg/pdfcpu/read.go#L576

fancycode:
Fix #1012

### DIFF
--- a/pkg/pdfcpu/read.go
+++ b/pkg/pdfcpu/read.go
@@ -463,10 +463,10 @@ func parseObjectStream(c context.Context, osd *types.ObjectStreamDict) error {
 	return nil
 }
 
-func createXRefTableEntryFromXRefStream(entry byte, objNr int, c2, c3, offExtra int64, objStreams types.IntSet) model.XRefTableEntry {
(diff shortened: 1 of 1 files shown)


hhrutter:

PR #1019, opened 2024-12-02
Support float page rotation with zero fractional part.

### LINKED ISSUE #1018: Parsing error if page rotation is float with all-zero fraction

Please ensure the following:

Ubuntu Linux 20.04

Page resources are stored with dictionaries like this:
```
<<
  /Resources <<
    /XObject <<
      /F 111 0 R
      /FFT0 222 0 R
    >>
  >>
  /MediaBox[ 0 0 2400 3400]
  /Type /Page
  /Parent 333 0 R
  /Contents 444 0 R
  /Rotate 90.0
  /CropBox [ 0 0 2400 3400]
>>
```

During parsing, the `/Rotate` entry is read as integer:
https://github.com/pdfcpu/pdfcpu/blob/ff51203026d797e891c145229251bd53cf5307dd/pkg/pdfcpu/model/xreftable.go#L1713-L1714

This is according to the standard which defines the value as `integer`, however there seem to be some writers that store the value as floating point number with an all-zero fractional part.

The `integer` datatype is defined as "An integer **SHALL** be written as one or more decimal digits optionally preceded by a sign." (emphasis mine), so imho it should be fine to also support cases like the above and reject all values that have a non-zero fractional component.

fancycode:
Fixes #1018

### DIFF
--- a/pkg/pdfcpu/model/xreftable.go
+++ b/pkg/pdfcpu/model/xreftable.go
@@ -22,6 +22,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io"
+	"math"
 	"os"
 	"path"
 	"sort"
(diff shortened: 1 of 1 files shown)


hhrutter:
Hi!

Why should we round within epsilon?
What's the motivation behind?
Do you see files with such weird rotation values?

fancycode:
The problem with floating point is that not all numbers can be specified exactly with the underlying internal representation, e.g. `0.15 + 0.15` is `0.3` but `0.1 + 0.2` could be `0.30000000000000004`.

While this might not happen for the rotation values, I figured better be on the safe side and round with epsilon.

hhrutter:
Unless you have input files that use a rotation values of 0.30000000000000004 and such I am against rounding.

The issue here is while all may be true what might happen any PDF Writer has to obey the spec and come up with an int rotation value.

fancycode:
Updated to remove the epsilon rounding and only support float values for relaxed validation.

PR #1021, opened 2024-12-02
Handle missing page tree node type for pages.

### LINKED ISSUE #1020: nil pointer dereference if page tree node has no "Type"

Please ensure the following:

Ubuntu Linux 20.04

Most of the pages nodes have no `Type` entry, causing a nil-pointer deference here:
https://github.com/pdfcpu/pdfcpu/blob/ff51203026d797e891c145229251bd53cf5307dd/pkg/pdfcpu/model/xreftable.go#L1916

I suggest updating the node type to `Pages` if it has `Count` and `Kids` (this fixes the crash.

fancycode:
Fix #1020

### DIFF
--- a/pkg/pdfcpu/model/xreftable.go
+++ b/pkg/pdfcpu/model/xreftable.go
@@ -1943,6 +1943,21 @@ func (xRefTable *XRefTable) processPageTreeForPageDict(root *types.IndirectRef, 
 				return xRefTable.processPageTreeForPageDict(&indRef, pAttrs, p, page, consolidateRes)
 			}
 
+		case "Template":
+			if xRefTable.ValidationMode == ValidationStrict {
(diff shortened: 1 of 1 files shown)


hhrutter:
If we are using strict validation and we encounter a page dict without type we need to return an error rather than ignoring the page dict.

fancycode:
Ok, what about unsupported values? The current code currently ignores these:
https://github.com/pdfcpu/pdfcpu/blob/56bed9fbf1556513f3f33946036768ad51c1bde4/pkg/pdfcpu/model/xreftable.go#L1916-L1934

Should I return an error there in a `default:` statement, too?

hhrutter:
The possible types are *Pages*, *Page* and *Template*.
Template is unsupported at the moment.
Therefore for page dict type *Template* we should return a corresponding error.
For any other than the 3 predefined types, let's return a "missing pages dict type" error.

PR #1027, opened 2024-12-05
Don't decrypt "Contents" of "DocTimeStamp" signatures.

### LINKED ISSUE #1026: Handle "DocTimeStamp" signatures in encryption / decryption.
Please ensure the following:

Ubuntu Linux 20.04

Affected are all signed and encrypted documents that support PDF Advanced Electronic Signature Profiles with long term validation as defined by ETSI TS 102 778-4.

-----

Similar to #794, there also are signature dictionaries with a `Type` of `DocTimeStamp`. In these dictionaries, the `Contents` also must not be decrypted / encrypted.

The better solution than checking the different types would be to "know" which dictionaries are "signature dictionaries", i.e. dictionaries referenced from the `/V` entry of `/Sig` annotations. However this currently is not easily possible as objects are decrypted / encrypted individually without having a context where they belong to / where they are referenced from.

So as an intermediate workaround, an additional check for `DocTimeStamp` should be added. PR incoming...

fancycode:
Fix #1026

### DIFF
--- a/pkg/pdfcpu/crypto.go
+++ b/pkg/pdfcpu/crypto.go
@@ -1280,7 +1280,7 @@ func encryptDict(d types.Dict, objNr, genNr int, key []byte, needAES bool, r int
 		ft = d["Type"]
 	}
 	if ft != nil {
-		if ftv, ok := ft.(types.Name); ok && ftv == "Sig" {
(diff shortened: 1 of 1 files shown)


hhrutter:
Thank you for your contribution!!!

PR #1028, opened 2024-12-08
Bump GitHub actions to latest versions

deining:
This PR bumps GitHub workflow actions to their latest versions. It also fixes warnings as e.g. seen [here](https://github.com/pdfcpu/pdfcpu/actions/runs/12121154268).

```
Restore cache failed: Dependencies file is not found in /home/runner/work/pdfcpu/pdfcpu.
Supported file pattern: go.sum
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+version: 2
+updates:
+  # Enable version updates for GitHub action workflows
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    # Check for updates to GitHub Actions once per week
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
+  # Enable version updates for Go modules
+  - package-ecosystem: gomod
+    directory: /
+    # Check for updates to Go modules once per week
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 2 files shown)


fancycode:
Might be a good idea to setup @dependabot to handle future dependency updates automatically.

hhrutter:
Excellent idea.

deining:
> Excellent idea.

Just added a commit that implements automatic update checking via dependabot.

hhrutter:
I would like the update interval to be weekly without a specific time.
If you also leave test.yml untouched I am happy to accept this PR.
The dependabot PR will take care of the rest.

deining:
> I would like the update interval to be weekly without a specific time.

I just changed `dependabot.yml` as requested.

> If you also leave test.yml untouched

Reverted `test.yml`, but swapped checkout and go setup action in order to avoid the warning `Restore cache failed`.

> I am happy to accept this PR.

Thanks for merging!

hhrutter (commented):
Go modules update interval is still daily.

deining:
> Go modules update interval is still daily.

Oops, sorry. This is fixed now.

hhrutter:
Thanks!

Issue #1043, opened 2024-12-16
wkhtmltopdf generate pdf  COMPATIBLE issue


frankz61:
when i use pdfcpu to merge files, i found a compatible problem that wkhtmltopdf generate file.

and it seem normal in adobe acrobat, 
![image](https://github.com/user-attachments/assets/28021729-98a6-4435-aed5-06b4c1a69bd1)

when program execute validateDestinationArrayLength, the invaild length: 10 error is returned.
![image](https://github.com/user-attachments/assets/4621bf97-6d79-41bc-8d93-d2eeb979a183)

i checked the pdf file, found it has mutilple array. 

![image](https://github.com/user-attachments/assets/f19b7cc7-f64d-47ed-a7d9-67b672646a2e)

so, maybe add length 10 vaildate will be fine.

hhrutter:
This is a PDF specification violation.
The syntax for a destination array of type XYZ is defined having 5 entries:

<img width="1142" alt="Screenshot 2024-12-16 at 18 16 30" src="https://github.com/user-attachments/assets/b358f36e-a592-45e5-b006-e7efdeef0081" />

frankz61:
that`s right,
It should be a wkhtmltopdf program issue, because i see a wkhtmltopdf generate file that have XYZ 25 object length.
![image](https://github.com/user-attachments/assets/d43d90f7-6869-45a8-a664-5921961e84ff)

pdf format really has so many difficult implements.

it is my code to solve this, before check length to solve this error data.
	o, err := validateDestinationArrayFirstElement(xRefTable, a)
	if err != nil || o == nil {
		return err
	}

	// XYZ - 5 elements compatible, in wkhtmltopdf maybe multiple times. look like [0 /XYZ 28.5000000 753.500000 0 0 /XYZ 28.5000000 753.500000 0]
	if a[1] == types.Name("XYZ") && len(a) > 5 {
		a = a[:5]
	}

hhrutter:
Closing this.

Checkout #1046 for an explanation.

PR #1046, opened 2024-12-17
solve parse wkhtmltopdf XYZ multiple object issue and datetime format…

frankz61:
This PR fixes an issue addressed in #1043.

### DIFF
--- a/pkg/pdfcpu/types/date.go
+++ b/pkg/pdfcpu/types/date.go
@@ -54,6 +54,10 @@ func prevalidateDate(s string, relaxed bool) (string, bool) {
 	// Remove trailing 0x00
 	s = strings.TrimRight(s, "\x00")
 
+	// Remove "()", dateformat just like (D:20060102150405-07'00')
+	s = strings.TrimPrefix(s, "(")
+	s = strings.TrimSuffix(s, ")")
+
 	if relaxed {
 		// Accept missing "D:" prefix.
 		// "YYYY" is mandatory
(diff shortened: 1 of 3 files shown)


hhrutter:
Hello!

Since this PR relates to Date objects generated by this has nothing to do with #1043.

I am happy to look into this, can you share a simple file that demonstrates this issue during validation?

If wkhtmltopdf generates Dates that are not conform to the PDF spec then please file an issue with that project which is also hosted on Github.

-> https://github.com/wkhtmltopdf/wkhtmltopdf

frankz61:
Hi, hhrutter. Nice to see your reply, 
the problem of the XYZ multiple object, i`m sure the file is generate by wkhtmltopdf.
the file is 
image
[1.pdf](https://github.com/user-attachments/files/18217144/1.pdf)
image
[2.pdf](https://github.com/user-attachments/files/18217145/2.pdf)

the second problem of the datetime format, it might not be generate by wkhtmltopdf, i see the file information like this,
image
it means Application : maker, PDF Producer: producer.  i`m not sure, which app generated it.
this file is 
image
[3.pdf](https://github.com/user-attachments/files/18217154/3.pdf)

hhrutter:
https://github.com/wkhtmltopdf/wkhtmltopdf is an archived repo in Alpha with 1.3k open issues.

⚡  I recommend staying away from wkhtmltopdf if you are planning any post processing⚡ 

This illustrates one of the big issues when processing with real world PDFs.
Folks will always prefer to use free tools grabbed quickly from Github  for PDF generation
and then all PDF processors out there are left to digest the resulting malformed PDFs.

Only because you can open a PDF generated with such a tool does not necessarily mean the produced file conforms to the PDF spec which is a definite assumption for further processing.

Closing this.

frankz61:
I understand you.
Maybe I was a client programmer, so the different problems related to the environment always confuse me. And I want to solve them.
Just like this: different code and different implementations, and not follow standard well.
Making the original code clean and regular is important.
I respect you.

Issue #437, opened 2022-02-17
dict=Link entry=QuadPoints: unsupported in version 1.4\nThis file could be PDF/A compliant but pdfcpu only supports versions <= PDF V1.7


mpseay:
Hey pdfcpu!
I've got a PDF here unfortunately I can't share. But, attempting to merge/validate it yields:

`dict=Link entry=QuadPoints: unsupported in version 1.4\nThis file could be PDF/A compliant but pdfcpu only supports versions <= PDF V1.7\n`

After further investigation, it looks like the library is explicit about supporting Quadpoints with v1.6. However, I was able to perform both of these actions with a slight modification although I'm not entirely sure of the impact:

```
diff --git a/pkg/pdfcpu/validate/annotations.go b/pkg/pdfcpu/validate/annotations.go
index 0aed620..b63a01a 100644
--- a/pkg/pdfcpu/validate/annotations.go
+++ b/pkg/pdfcpu/validate/annotations.go
@@ -330,7 +330,7 @@ func validateAnnotationDictLink(xRefTable *pdf.XRefTable, d pdf.Dict, dictName s
        }

        // QuadPoints, optional, number array, len= a multiple of 8, since V1.6
-       _, err = validateNumberArrayEntry(xRefTable, d, dictName, "QuadPoints", OPTIONAL, pdf.V16, func(a pdf.Array) bool { return len(a)%8 == 0 })
+       _, err = validateNumberArrayEntry(xRefTable, d, dictName, "QuadPoints", OPTIONAL, pdf.V14, func(a pdf.Array) bool { return len(a)%8 == 0 })
        if err != nil {
```

thoughts?

hhrutter:
Hello!

We can extend relaxed validation to take this spec violation into account.

mpseay:
Awesome-sauce! Much appreciated!

mpseay:
@endigo Fix looks good to me. I'm currently running

`github.com/pdfcpu/pdfcpu v0.3.14-0.20220309084027-515c2eff91ec`

You can install the latest commit(5 days ago) `go get github.com/pdfcpu/pdfcpu@8f9a93cb9b537e4de850082e70a6fffd8e18512a`

Hope that helps!

hhrutter:
Unfortunately I am unable to cut a release at this point. Please go get and install the current master branch like suggested above.

Thank you for using pdfcpu

PR #1063, opened 2025-01-06
Fix #1058 Use relaxed validation for dict=Square entry=RD

### LINKED ISSUE #1058: Error dict=Square entry=RD: unsupported in version 1.4
PDF Document Version: v1.4
pdfcpu Version: `v0.9.1`
OS: MacOS and on Dockre deployment of the below code (Alpine base image).

The following code snippet fails with an error:

```golang
	conf := pdfmodel.NewDefaultConfiguration()
	pageCount, err := pdfmeta.PageCount(pdfFile, conf)
	if err != nil {
		panic(err)
	}
```

(`pdfFile` is the `*os.File` pointing to the pdf document)

The error:

```
Error dict=Square entry=RD: unsupported in version 1.4
```

vishal-at:
Fixes #1058 

We have a PDF (can't share as it contains sensitive info) which while attempting to validate gives `Error dict=Square entry=RD: unsupported in version 1.4`  Library is explicitly validating against version 1.5 for this. 

I went through https://github.com/pdfcpu/pdfcpu/commit/515c2eff91ecbe5a88147133b581bea68c352526 for #437 
and added a similar change here.

### DIFF
--- a/pkg/pdfcpu/validate/annotation.go
+++ b/pkg/pdfcpu/validate/annotation.go
@@ -672,6 +672,11 @@ func validateAnnotationDictCircleOrSquare(xRefTable *model.XRefTable, d types.Di
 		return err
 	}
 
+	sinceVersion = model.V15
+	if xRefTable.ValidationMode == model.ValidationRelaxed {
+		sinceVersion = model.V14
+	}
+
 	// RD, optional, rectangle, since V1.5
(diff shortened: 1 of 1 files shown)


hhrutter (commented):
This should be:
This should be:
sinceVersion = model.V15
if xRefTable.ValidationMode == model.ValidationRelaxed {
	sinceVersion = model.V14
}

hhrutter:
Thanks for the contribution!

PR #1066, opened 2025-01-08
Avoid duplicate keyword search

### LINKED ISSUE #1057: Improve the speed of reading basic PDF information
Sometimes I just want to read the number of pages in a PDF.
A 1500-page PDF takes more than 10 minutes to read using pdfcpu, even if the optimization option is turned off.

fancycode:
Fixes #1057

The example file contains a huge object with this content:
```
210891 0 obj
<</Limits [(node00000002) (node00275919)]
/Names [(node00000002) 8591 0 R (node00000006) 190838 0 R (node00000007) 66313 0 R (node00000008) 24685 0 R (node00000009) 208328 0 R (node00000010) 163561 0 R (node00000011) 27134 0 R (node00000012) 126131 0 R (node00000013) 9895 0 R (node00000014) 101713 0 R (node00000015) 64243 0 R (node00000016) 118039 0 R (node00000017)
... continues for more than 5 MBytes ...
(node00275919) 209354 0 R]>>
endobj
```

This blows up the parsing code in https://github.com/pdfcpu/pdfcpu/blob/4e0e0db8c786122a8efbe5fb71c070810a4aca1d/pkg/pdfcpu/model/parse.go#L1243
Here the keywords are searched lots of times after the start is advanced for each string literal inside the array (e.g. `(node00000002)`, etc.).

The PR fixes this by searching only once for the keywords and then adjusting the position if strings / comments are found before them. In my tests the example file now parses in about 3 seconds.

### DIFF
--- a/pkg/pdfcpu/read.go
+++ b/pkg/pdfcpu/read.go
@@ -1738,7 +1738,7 @@ func buffer(c context.Context, rd io.Reader) (buf []byte, endInd int, streamInd 
 		growSize = min(growSize*2, maximumBufSize)
 		line := string(buf)
 
-		endInd, streamInd, err = model.DetectKeywords(line)
+		endInd, streamInd, err = model.DetectKeywordsWithContext(c, line)
 		if err != nil {
 			return nil, 0, 0, 0, err
 		}
--- a/pkg/pdfcpu/model/parse.go
+++ b/pkg/pdfcpu/model/parse.go
@@ -1168,7 +1168,7 @@ func posFloor(pos1, pos2 int) int {
 		return pos2
 	}
 	if pos1 < pos2 {
-		return pos2
+		return pos1
 	}
 	if pos2 < 0 {
 		return pos1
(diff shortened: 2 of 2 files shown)


hhrutter:
Sounds very cool..
Let me double check your patch of this critical code section.

pkg/pdfcpu/read.go
```diff
@@ -1721,7 +1721,7 @@ func buffer(c context.Context, rd io.Reader) (buf []byte, endInd int, streamInd
 		growSize = min(growSize*2, maximumBufSize)
 		line := string(buf)
 
-		endInd, streamInd, err = model.DetectKeywords(line)
+		endInd, streamInd, err = model.DetectKeywordsWithContext(c, line)
```
hhrutter, line 1724: Why don't we leave the original method name DetectKeywords and pass in as 2nd parm the context from the caller. does that make any sense?

hhrutter (commented):
There is a(n old) bug in posFloor:

```
if pos1 < pos2 {
    return pos2
}
```

should be 
```
if pos1 < pos2 {
    return pos1
}
```

hhrutter:
Happy to merge if you can take care of these 2 issues.

Issue #1014, opened 2024-11-26
Extract pages corrupts form


mdmcconnell:
I am running pdfcpu on the command line under wsl Ubuntu.  This behaviour is occuring under the most recent commit b9b3d0fca12d.

Using "pdfcpu extract -mode page f1040se.pdf" out on the first attached file yields the second attached.  The extracted page is no longer a valid form in some way.  First, Acrobat has the following error:
![image](https://github.com/user-attachments/assets/5eab2e6f-36af-4690-9858-80fa5207c752)

But more interestingly, pressing tab from any form field causes the reader to hang.
[f1040se.pdf](https://github.com/user-attachments/files/17919597/f1040se.pdf)
[f1040sep1.pdf](https://github.com/user-attachments/files/17919598/f1040sep1.pdf)

hhrutter:
Well, seems like you are ripping your form apart..
Essentially what you want is `form splitting`, that's not implemented.
I also doubt it ever will be since there maybe form field dependencies between pages of your form containing fieds.

mdmcconnell:
Yes, that's a good point - I did not know that this is possible, and have not seen it, but if there was some active element to the form and a cross dependency between fields, it would clearly be impossible separate those fields.  In the simple case of no dependency, I think it should be possible.  Or pdfcpu could throw an error instead of creating a (presumably invalid) PDF that crashes Acrobat.

hhrutter:
Closed due to inactivity.

PR #1075, opened 2025-01-16
Two questions on migrateObject

mdmcconnell:
It looks like the simplest way to ask code questions is by creating a pull request with comments, so that is what I have done.  I don't expect to merge it - that would be pointless.

### DIFF
--- a/pkg/pdfcpu/migrate.go
+++ b/pkg/pdfcpu/migrate.go
@@ -48,6 +48,7 @@ func migrateObject(o types.Object, ctxSource, ctxDest *model.Context, migrated m
 	case types.IndirectRef:
 		objNr := o.ObjectNumber.Value()
 		if migrated[objNr] > 0 {
+			// Can this clause ever change migrated[objNr]?  If it is already migrated, shouldn't it have the new objNr?
 			o.ObjectNumber = types.Integer(migrated[objNr])
 			return o, nil
 		}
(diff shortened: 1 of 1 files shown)


hhrutter:
Sorry, I am busy dealing with resolving issues and I don't really have the time for explaining the code base right now.
If this relates to an issue while using pdfcpu please create one and lay out your specific problem and use case ideally with an attached file for analysis.

Thank you for understanding!

mdmcconnell:
This relates to trying to solve https://github.com/pdfcpu/pdfcpu/issues/1014 .  In splitting the form, pdfcpu flattens the field tree, and something goes wrong.  I suspect a circular reference, since Acrobat hangs.

hhrutter:
Splitting a PDF that represents a PDF form is just not supported and probably never will be.
See my comments at #1014.

The result of splitting a PDF form is a sequence of pages  where each page containing inputfields is a completely new form.
If anybody comes across a tool that manages to do this please let me know.

mdmcconnell:
I'm sure there are complex cases that will be difficult to handle, but the simple case of "a page with some boxes on it for the user to fill" is really not so difficult.  pdfcpu attempts to do it, and is nearly successful but fails with a subtle error - which the program itself then manages to fix during pdfcpu form fill!  qpdf also creates a working split form, though I suspect it is leaving all form fields in every output file (such that many are not displayed).

Issue #1118, opened 2025-04-02
Handle escape characters in ModDate


GwendalSuperG:
We are using pdfcpu to handles eInvoicing with Factur-X. The Factur-X format add an XML as an attachments to the PDF invoice to allow easy processing.

We encounter an issue with some samples PDF provided by Factur-X https://fnfe-mpe.org/factur-x/
The ModDate is invalid:
```
$ pdfcpu attachments list Facture_F20220023-LE_FOURNISSEUR-POUR-LE_CLIENT_MINIMUM.pdf 
validating URIs..

pdfcpu: invalid date ModDate
```

This is because the ModDate contains escaped characters for example:
D\07220200707112545\05300\04700\047

It can be solved by adding an Unqote call on the DateTime function:

```
// DateTime decodes s into a time.Time.
func DateTime(s string, relaxed bool) (time.Time, bool) {
	// 7.9.4 Dates
	// (D:YYYYMMDDHHmmSSOHH'mm)
	var d time.Time
        // Add an unquote
        var err error
	s, err = strconv.Unquote(fmt.Sprintf("\"%s\"", s))
	if err != nil {
		return d, false
	}


	var ok bool
	s, ok = prevalidateDate(s, relaxed)
	if !ok {
		return d, false
	}
...
```

You can find a PDF file with this issue and the full verbose report

[verbose.txt](https://github.com/user-attachments/files/19564034/verbose.txt)
[Facture_F20220023-LE_FOURNISSEUR-POUR-LE_CLIENT_MINIMUM.pdf](https://github.com/user-attachments/files/19563961/Facture_F20220023-LE_FOURNISSEUR-POUR-LE_CLIENT_MINIMUM.pdf)

If you are ok with this I can submit a PR

hhrutter:
Thanks for reporting this.

There is a code freeze in effect as part of the preparation for a new release.

milktean:
When will a new release be published?

hhrutter:
over the weekend..

hhrutter:
This is fixed with the latest commit!

I apologize this somehow did not make it into the release.
The last days were hectic to say the least and so I am pushing a fix straight through.

Your PR suggests modifying `date.go`, but `date.go` is not the culprit here.
Instead date strings need to be extracted properly.

PR #1140, opened 2025-04-25
Fix https://github.com/pdfcpu/pdfcpu/issues/1118

GwendalSuperG:
This fix error when the ModDate contains escaped characters. There are PDF with this, especially regarding the e-Invoicing reform in France
A pdf containing this and detail log error are attached to this issue: https://github.com/pdfcpu/pdfcpu/issues/1118


hhrutter:
Thanks for this, but I am going to close this.
See my reasoning at #1118

Issue #1149, opened 2025-05-07
Panic during pageEntryThumb:  assignment to entry in nil map


tgruben-circuit:
looks like a map needs allocating 

### Thank you for submitting a possible bug!
```
panic:

[cookbook.pdf](https://github.com/user-attachments/files/20085412/cookbook.pdf)

goroutine 1 [running]:
github.com/pdfcpu/pdfcpu/pkg/pdfcpu/validate.validatePageEntryThumb(0x1400039e908, 0x140000edc80, 0x78?, 0x4?)
        /Users/toddgruben/go/pkg/mod/github.com/pdfcpu/pdfcpu@v0.10.2/pkg/pdfcpu/validate/page.go:311 +0xd4
github.com/pdfcpu/pdfcpu/pkg/pdfcpu/validate.validatePageDict(0x1400039e908, 0x140000edc80, 0x140006c50c0?, 0x0, 0x1)
        /Users/toddgruben/go/pkg/mod/github.com/pdfcpu/pdfcpu@v0.10.2/pkg/pdfcpu/validate/page.go:889 +0x3f4
github.com/pdfcpu/pdfcpu/pkg/pdfcpu/validate.processPagesKids(0x1400039e908, {0x14000716800?, 0x1?, 0x0?}, 0x2, 0x0, 0x1, 0x140000e7870)
        /Users/toddgruben/go/pkg/mod/github.com/pdfcpu/pdfcpu@v0.10.2/pkg/pdfcpu/validate/page.go:1027 +0x2dc
github.com/pdfcpu/pdfcpu/pkg/pdfcpu/validate.validatePagesDict(0x1400039e908, 0x140000ecf60, 0x2, 0x0, 0x0, 0x140000e7870)
        /Users/toddgruben/go/pkg/mod/github.com/pdfcpu/pdfcpu@v0.10.2/pkg/pdfcpu/validate/page.go:1062 +0x88
github.com/pdfcpu/pdfcpu/pkg/pdfcpu/validate.validatePages(0x1400039e908, 0x140000ecf30)
        /Users/toddgruben/go/pkg/mod/github.com/pdfcpu/pdfcpu@v0.10.2/pkg/pdfcpu/validate/page.go:1179 +0x238
github.com/pdfcpu/pdfcpu/pkg/pdfcpu/validate.validateRootObject(0x140000ed170)
[...]

PR #1151, opened 2025-05-07
protect against uninitialized map in validatePageEntryThumb

tgruben-circuit:
The issue was that somewhere in the codebase, an XRefTable instance is being created without using the newXRefTable function that properly initializes all map fields, including PageThumbs. The fix I've provided checks if PageThumbs is nil and initializes it if needed, ensuring that the code can handle cases where the map wasn't properly initialized.

 Fixes # 1149


hhrutter:
The related issue is important but this is not a proper fix.
See #1149

Next time please open an issue and we discuss and decide further steps.

No premature PRs.
Thank you!

PR #1153, opened 2025-05-10
Introduce `WritableFS`

olekukonko:
The current Extraction functions make an assumption that writers will always write to a local directory. This limits flexibility and prevents integration with various storage backends like:
- Block storage systems
- HTTP APIs
- In-memory storage
- Temporary file systems
- etc

**Proposed Solution:**
Introduce a `WritableFS` interface that abstracts file system operations, allowing for multiple storage implementations. This provides:

```go
// WritableFS defines an interface for a writable file system or directory-like storage.
type WritableFS interface {
    // Base returns the base path or identifier of the file system (e.g., directory path or "memory").
    Base() string

    // WriteFile writes data to the named file, creating it if necessary.
    // If the file exists, it may overwrite it (implementation-dependent).
    // The perm parameter specifies the file permissions (e.g., 0644).
    WriteFile(name string, data []byte, perm fs.FileMode) error

    // Mkdir creates a directory named path, along with any necessary parents.
    // The perm parameter specifies the directory permissions (e.g., 0755).
    Mkdir(path string, perm fs.FileMode) error
}
```

1. Better separation of concerns
2. Flexibility to use different storage backends
3. Cleaner API surface
4. Improved testability with in-memory implementations

**Key Changes:**
1. Added `WritableFS`
2. Implemented Two concrete types:
   - `OsFS` for local file system operations
   - `MemFS` for in-memory storage (useful for testing)

**Example Usage:**

```go
// Using local filesystem
osFs := NewOsFS("/output/dir")

// Using in-memory storage (for testing)
memFs := NewMemFS()
```

```go
// Normal way:
ExtractContent(inFile, "/output/dir", pages, conf)

// Using WritableFS:
ExtractContentFS(inFile, NewOsFS("/output/dir"), pages, conf)
```

**Benefits:**
- Backward compatible (can still use local file system)
- Enables new storage integrations without changing core logic
- Makes code more testable
- Follows Go's interface-oriented design principles
- Opens possibilities for future storage implementations (S3, GCS, etc.)

This is a production code hence Baby steps are required, hence i wanted to start with ExtractContentFS

### DIFF
--- a/pkg/api/test/extract_test.go
+++ b/pkg/api/test/extract_test.go
@@ -318,3 +318,21 @@ func TestExtractMetadataLowLevel(t *testing.T) {
 			md.ObjNr, md.ParentObjNr, md.ParentType, string(bb))
 	}
 }
+
+
+func TestExtractContentCommandFS(t *testing.T) {
+	msg := "TestExtractContentCommandFS"
+
+	// Extract content of all pages into outDir.
+	inFile := filepath.Join(inDir, "5116.DCT_Filter.pdf")
+
+	fs, err := os.Open(inFile)
+	if err != nil {
+		t.Fatalf("%s %s: %v\n", msg, inFile, err)
+	}
+
+	fsOut := api.NewOsFS(outDir)
+	if err := api.ExtractContentFS(fs, fsOut, nil, nil); err != nil {
+		t.Fatalf("%s %s: %v\n", msg, inFile, err)
+	}
+}
--- a/pkg/api/writeable.go
+++ b/pkg/api/writeable.go
@@ -1,0 +1,226 @@
+package api
+
+import (
+	"fmt"
+	"github.com/pdfcpu/pdfcpu/pkg/log"
+	pdf "github.com/pdfcpu/pdfcpu/pkg/pdfcpu"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"time"
+)
+
+// WritableFS defines an interface for a writable file system or directory-like storage.
+type WritableFS interface {
+	// Base returns the base path or identifier of the file system (e.g., directory path or "memory").
+	Base() string
+
+	// WriteFile writes data to the named file, creating it if necessary.
+	// If the file exists, it may overwrite it (implementation-dependent).
+	// The perm parameter specifies the file permissions (e.g., 0644).
+	WriteFile(name string, data []byte, perm fs.FileMode) error
+
+	// Mkdir creates a directory named path, along with any necessary parents.
+	// The perm parameter specifies the directory permissions (e.g., 0755).
+	Mkdir(path string, perm fs.FileMode) error
+}
+
+// OsFS is a WritableFS implementation for the local file system.
+type OsFS struct {
+	BasePath string // BasePath is the root directory for file operations.
+}
+
+// NewOsFS creates a new OsFS with the specified base path.
+func NewOsFS(basePath string) *OsFS {
+	return &OsFS{BasePath: basePath}
+}
+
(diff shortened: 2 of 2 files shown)


hhrutter:
Please create a feature request  issue.
PRs are the result of issue discussions.
Thank you!

olekukonko:
### **What Problem Does `WritableFS` Solve?**
The core idea is **abstracting where extracted PDF content gets written**. Instead of hardcoding filesystem operations, the PDF extraction logic delegates writing to an interface (`WritableFS`). This enables:
1. **Writing to disk** (via `OsFS`).  
2. **Writing to memory** (via `MemFS`, useful for tests).  
3. **Writing to stdout or pipes** (see below).  
4. **Future backends** (e.g., cloud storage, ZIP files, databases).  


### **Example**
#### 1. **Support for Pipes/Stdout (Your Example)**
The comment you referenced highlights a real pain point: **temporary files break clean piping** (e.g., `pdfcpu | jq | pdfcpu`). With `WritableFS`, you could add a `StdoutFS` implementation:
```go
type StdoutFS struct{} // Implements WritableFS

func (s *StdoutFS) WriteFile(name string, data []byte, _ fs.FileMode) error {
    _, err := os.Stdout.Write(data)
    return err
}

// Usage: ExtractContentFS(pdfData, &StdoutFS{}, ...) 
```
Now, the extracted PDF content streams directly to stdout—**no temp files needed**. This fits Unix-style pipelines perfectly.

#### 2. **Testability Without I/O Overhead**
`MemFS` lets you verify *what would be written* without touching the disk:
```go
func TestExtractContent(t *testing.T) {
    memFS := NewMemFS()
    err := ExtractContentFS(pdfReader, memFS, []string{"1"}, nil)
    // Check memFS.files["1_123.txt"] contains expected data
}
```
- **Faster tests**: No filesystem cleanup.  
- **More reliable**: No permission/disk space issues.  

#### 3. **Swappable Storage Backends**
The same extraction logic works across backends. For example, to add S3 support:
```go
type S3FS struct { Bucket string }

func (s *S3FS) WriteFile(name string, data []byte, _ fs.FileMode) error {
    return s3.Upload(s.Bucket, name, data)
}
```
No changes to `ExtractContentFS` are needed—just pass `S3FS` instead of `OsFS`.

#### 4. **Consistent Directory Handling**
Both `OsFS` and `MemFS`:
- Normalize paths (e.g., `clean/../path` → `path`).  
- Auto-create parent directories (no "directory not found" errors).  
- Respect file permissions (even if just in-memory for tests).  


### **Why This Design Matters**
- **Separation of concerns**: PDF extraction doesn’t care *where* data goes.  
- **Unix philosophy**: Enable pipes by supporting stdout (via `StdoutFS`).  
- **Future-proof**: New storage backends don’t require refactoring.  

For your specific use case (avoiding temp files in pipes), the missing piece is a `StdoutFS` implementation—which would be trivial to add thanks to `WritableFS`. The interface makes this *explicitly possible* without hacks.  

The only thing is that am not sure if ` _ fs.FileMode` is even needed

hhrutter:
Thanks @olekukonko for this proposal and for opening #1155 for the design discussion.

The underlying goal makes sense, especially around piping/stdout and avoiding unnecessary temporary files. Since this PR was opened, we have been moving pdfcpu in a different architectural direction: consistent stream-based processing across the CLI and Go API, using reader/writer style boundaries instead of introducing a separate writable filesystem abstraction.

Because of that, `WritableFS` no longer fits the direction we are taking here, and this PR is also stale relative to the current codebase. I am going to close it for now.

Thanks again for taking the time to think this through and contribute.

Issue #1211, opened 2025-08-15
Add wrapping and justification to form text field filling


mdmcconnell:
My application often fills multi-line fields.  Text can overrun a single line, leading to this sort of output:

<img width="305" height="47" alt="Image" src="https://github.com/user-attachments/assets/0bc12ae9-956d-4e83-bca2-cb23f3e7fa01" />

The text is justified to the top of the field, but truncated at the right edge. I do not have a good way to prevent that at run time, so I suggest two possibilities:

1 - allow a FillDetails function to specify behaviour - modify function signature from

`func(id, name string, fieldType FieldType, format DataFormat) ([]string, bool, bool)`
to
```
type WrapBehaviour struct {
	Wrap              bool // Wrap the text if it it does not fit on single line of multi-line field
	Error             bool // Generate an error if the text overruns after applying all wrapping behaviours
	VeticalJustify    int
	HorizontalJustify int
	Compress          bool // Compress text horizontally as necessary/reasonable to fit the field
}

func(id, name string, fieldType FieldType, format DataFormat) ([]string, bool, bool, WrapBehaviour)
```

2 - Give FillDetails information it needs to determine wrapping for itself
`func(id, name string, fieldType FieldType, format DataFormat, boundingbox, fontinfo) ([]string, bool, bool, fontinfo)`

There would be follow-on changes to allow for these modifications when filling from a JSON or CSV file, like reading WrapBehaviour from JSON input.

mdmcconnell:
I think default behaviour is meant to be wrapping the text field, and a reader is responsible for doing this when it receives input into a field.  The two readers I am using do not automatically wrap the the text that has been programatically filled.  I think pdfcpu is generating appearance streams for the text.  I do not know if that should include layout as far a text wrapping - it seems the readers are expecting that pre-filled form text will have its appearance totally defined.

mdmcconnell:
Setting ctx.NeedAppearances before writing solves this specific problem for Acrobat, by forcing the reader to format the form field text.  gs seems to ignore the flag, and if I disable appearance stream generation entirely, gs fails badly at formatting fields.  I think that, if pdfcpu is going to generate appearance streams for form fields, it can do more.

hhrutter:
Can you please share a simple form/pdf which when filled reproduces this behavior?

Which viewer are you using?

mdmcconnell:
[...]

PR #1212, opened 2025-08-16
Add word wrapping to form text field rendering

mdmcconnell:
This PR partially deals with issue #1211 by adding word wrapping to text field rendering.  Also, no longer wraps fields which are not multiline, by converting newlines into spaces.  Unit tests for wrapping are included.


hhrutter:

PR #1239, opened 2025-10-03
Fix #1226: Correct typo in calcBorder()

### LINKED ISSUE #1226: Possible typo in calcBorder()
A static analysis tool suggested that `bb` should be `bbb` in the following code:
https://github.com/pdfcpu/pdfcpu/blob/master/pkg/pdfcpu/primitives/content.go#L731
I admit lacking expertise to confirm it, but I see no logical reason to treat the right border differently.
Similar sibling functions follow the same pattern and treat all margins/paddings/etc uniformly.

fank:
## Summary
- Fixed typo in `calcBorder()` where `Right` region was incorrectly passed `bb` instead of `bbb`
- This aligns with the pattern used in `calcMargin()` and `calcPadding()` functions
- Ensures Right border receives merged border configuration consistently with Left, Top, and Bottom

## Test plan
- Existing tests should pass
- The fix ensures consistent border handling across all regions

Fixes #1226

### DIFF
--- a/pkg/pdfcpu/primitives/content.go
+++ b/pkg/pdfcpu/primitives/content.go
@@ -728,7 +728,7 @@ func (c *Content) calcBorder(bb map[string]*Border) {
 	if c.Regions != nil {
 		if c.Regions.horizontal {
 			c.Regions.Left.calcBorder(bbb)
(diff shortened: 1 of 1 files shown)


fank:

hhrutter:

PR #1240, opened 2025-10-03
Fix #1229: Handle choice fields with missing Opt entry

### LINKED ISSUE #1229: "corrupt form field" when a form contains a listbox with no options
Using Adobe Acrobat, you can create a PDF document containing a ListBox form field with no options. The attached file, LISTBOX.pdf, is an example of such a document.

When running the following command with pdfcpu:

`pdfcpu form list LISTBOX.pdf`

the tool fails to parse the full PDF and returns the following error:
`corrupt form field: missing entry "Opt"`

As a result, the document cannot be fully accessed or processed.

Expected behavior:
pdfcpu should not crash or terminate parsing when encountering a ListBox with zero options. It should handle this edge case gracefully.

[LISTBOX.pdf](https://github.com/user-attachments/files/22299436/LISTBOX.pdf)

fank:
## Summary

Fixes #1229 by making the `Opt` entry optional for choice fields (ListBox and ComboBox), allowing pdfcpu to gracefully handle PDFs created by Adobe Acrobat that contain choice fields with no options.

## Problem

Adobe Acrobat allows creating ListBox and ComboBox fields with zero options, which results in PDFs without an `Opt` entry in the field dictionary. Previously, pdfcpu would fail with the error:

```
corrupt form field: missing entry "Opt"
```

This prevented users from listing, exporting, or processing such PDFs, even though they are valid according to the PDF specification (which marks `Opt` as optional).

## Solution

Changed all `parseOptions` calls from `REQUIRED` to `OPTIONAL` in:
- `pkg/pdfcpu/form/form.go` - `collectCh()` and `resetCh()` functions
- `pkg/pdfcpu/form/export.go` - `extractComboBox()` and `extractListBox()` functions  
- `pkg/pdfcpu/form/fill.go` - `fillChoiceField()` function

Each function now handles missing or empty options gracefully:
- **collectCh**: Skips setting options if none are present
- **resetCh**: Clears any existing values if no options available
- **export functions**: Exports fields with empty options array
- **fill functions**: Skips filling if no options available

## Testing

Tested with the provided `LISTBOX.pdf` file from the issue:

```
$ ./pdfcpu form list LISTBOX.pdf
/tmp/LISTBOX.pdf:

Pg L Field     │ Id  │ Name      
━━━━━━━━━━━━━━━┿━━━━━┿━━━━━━━━━━━
1    ListBox   │ 24  │ List Box1
```

The command now successfully lists the ListBox field without errors.

## Impact

This change improves compatibility with real-world PDF files created by Adobe Acrobat and follows the PDF specification's guidance that the `Opt` entry is optional for choice fields.

### DIFF
--- a/pkg/pdfcpu/form/export.go
+++ b/pkg/pdfcpu/form/export.go
@@ -374,15 +374,13 @@ func extractComboBox(xRefTable *model.XRefTable, page int, d types.Dict, id, nam
 		cb.Value = strings.TrimSpace(s)
 	}
 
-	opts, err := parseOptions(xRefTable, d, REQUIRED)
+	opts, err := parseOptions(xRefTable, d, OPTIONAL)
 	if err != nil {
 		return nil, err
 	}
-	if len(opts) == 0 {
-		return nil, errors.New("pdfcpu: combobox missing Opts")
+	if opts != nil {
+		cb.Options = opts
 	}
-
-	cb.Options = opts
 
 	return cb, nil
 }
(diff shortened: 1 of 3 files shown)


fank:

fank:
This more against the specs as written in the related issue, therefor not a solution.

PR #1241, opened 2025-10-03
Fix UTF-16 string parsing in balancedParenthesesPrefix

### LINKED ISSUE #1210: A normal PDF that cannot be opened normally with pdfcpu
```
pdfcpu info 'example.pdf'
dereferenceAndLoad: problem dereferencing object 2546: pdfcpu: parse: corrupt string literal, possibly unbalanced parenthesis
```
### some logs
`pdfcpu val -vv 'example.pdf'`
[debug.log](https://github.com/user-attachments/files/21791033/debug.log)

fank:
Fixes #1210

## Problem

The `balancedParenthesesPrefix` function was incorrectly counting raw `0x28` and `0x29` bytes that are part of UTF-16 encoded characters as parenthesis delimiters, causing parsing errors for PDF files with UTF-16 encoded strings.

For example, the Chinese character 用 (U+7528) is encoded as `0x75 0x28` in UTF-16BE. The `0x28` byte is not an opening parenthesis character—it's part of the character encoding.

## Root Cause

The parser was treating all `0x28` (opening paren) and `0x29` (closing paren) bytes as parenthesis delimiters, regardless of their context within UTF-16 encoded strings.

In UTF-16 encoded strings:
- An actual '(' character is `0x00 0x28` (UTF-16BE) or `0x28 0x00` (UTF-16LE)
- An actual ')' character is `0x00 0x29` (UTF-16BE) or `0x29 0x00` (UTF-16LE)
- Raw `0x28`/`0x29` bytes that are part of other character codes should be ignored

## Solution

Modified `balancedParenthesesPrefix` to:
1. Detect UTF-16 BOM markers (`0xFE 0xFF` for BE, `0xFF 0xFE` for LE)
2. Only count `0x28`/`0x29` bytes as parentheses when they represent actual UTF-16 encoded parenthesis characters
3. Properly handle the opening and closing string delimiters

## Testing

- Added comprehensive unit tests covering:
  - Basic parenthesis balancing
  - UTF-16BE encoded strings with embedded `0x28` bytes
  - UTF-16LE encoded strings
  - UTF-16 encoded actual parenthesis characters
  - Mixed ASCII in UTF-16 encoding
- Verified fix with the original PDF from issue #1210
- All existing tests pass (no regressions)

## Changes

- `pkg/pdfcpu/model/parse.go`: Updated `balancedParenthesesPrefix` function
- `pkg/pdfcpu/model/parse_test.go`: Added `TestBalancedParenthesesPrefix` with comprehensive test cases

### DIFF
--- a/pkg/pdfcpu/model/parse.go
+++ b/pkg/pdfcpu/model/parse.go
@@ -183,6 +183,12 @@ func balancedParenthesesPrefix(s string) int {
 	var j int
 	escaped := false
 
+	// Check for UTF-16 BOM after the opening '('
+	// UTF-16BE: 0xFE 0xFF, UTF-16LE: 0xFF 0xFE
+	isUTF16BE := len(s) > 2 && s[1] == '\xfe' && s[2] == '\xff'
+	isUTF16LE := len(s) > 2 && s[1] == '\xff' && s[2] == '\xfe'
+	isUTF16 := isUTF16BE || isUTF16LE
+
 	for i := 0; i < len(s); i++ {
 
 		c := s[i]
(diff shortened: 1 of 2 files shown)


fank:

petervwyatt:
Note that no version of PDF has ever defined support for UTF-16LE - only UTF-16BE, as per sub-clause 7.9.2.2.
But that doesn't mean you can't add it, but validation should fail such files (and especially for PDF/A, etc).

fank:
Thank you for the feedback @petervwyatt\! You're absolutely right - I'll update the implementation to only support UTF-16BE as per the PDF specification and remove UTF-16LE support. Let me fix that now.

fank:
Updated! I've removed UTF-16LE support from the implementation as it's not part of the PDF specification. The fix now only handles UTF-16BE (Big Endian) encoding, which aligns with:

1. The PDF specification (7.9.2.2)
2. The existing pdfcpu codebase (pkg/pdfcpu/types/utf16.go)

Changes in the latest commit:
- Removed UTF-16LE BOM detection
- Removed UTF-16LE parenthesis counting logic
- Removed UTF-16LE test cases
- Updated comments to clarify UTF-16BE-only support

All tests still pass with UTF-16BE support only.

hhrutter:
Already fixed.

PR #1242, opened 2025-10-03
Fix #1188: Automatically detect and preserve grayscale PNGs with alpha

### LINKED ISSUE #1188: Replacing an image with a grayscale with transparency one saves the image as DeviceRGB
If you replace an image in a pdf with a png that is grayscale + transparency, the image will be saved as DeviceRGB in the PDF

Image

[page_updated.pdf](https://github.com/user-attachments/files/20904843/page_updated.pdf)

This seems to be a problem with the go png library, as the color model that is returned by DecodeConfig is the same for ctGrayscaleAlpha and ctTrueColorAlpha (both return color.NRGBAModel )

fank:
## Summary

Fixes #1188 by automatically detecting grayscale PNG images (with or without alpha) and preserving them as DeviceGray color space instead of incorrectly converting them to DeviceRGB.

## Problem

When replacing images with PNG files that have grayscale+alpha (PNG color type 4), pdfcpu was saving them as DeviceRGB instead of DeviceGray, which:
- Resulted in 3x larger file sizes (RGB has 3 color components vs 1 for grayscale)
- Lost the original grayscale intent of the image
- Was inefficient for professional PDF workflows

The root cause was that Go's `image.Decode` converts both grayscale+alpha (color type 4) and RGB+alpha (color type 6) PNGs to the same `image.NRGBA` type, making them indistinguishable without inspecting the original PNG header.

## Solution

This PR implements automatic detection by:

1. **Reading PNG headers** - Added `getPNGColorType()` to parse PNG file headers and detect color type
2. **Automatic conversion** - When a grayscale PNG (color types 0 or 4) is detected, the decoded NRGBA image is converted to separate Gray and Alpha components
3. **Proper SMask extraction** - Alpha channel is properly extracted as a soft mask while using DeviceGray color space

## Implementation Details

- All changes are localized to `pkg/pdfcpu/model/image.go` as requested by maintainer feedback
- No new function parameters or API changes needed across the codebase
- Added helper functions:
  - `getPNGColorType()` - Reads PNG header to determine color type
  - `isGrayscaleNRGBA()` / `isGrayscaleNRGBA64()` - Validates that NRGBA pixel data is truly grayscale
  - `convertNRGBAToGrayAndAlpha()` / `convertNRGBA64ToGrayAndAlpha()` - Separates grayscale and alpha channels
  - `writeGrayWithAlphaImageBuf()` / `writeGray16WithAlphaImageBuf()` - Writes grayscale data with alpha mask

## Testing

- ✅ All existing tests pass
- ✅ Code builds without errors
- ✅ Tested with both 8-bit and 16-bit grayscale+alpha PNGs

## Comparison with PR #1213

This PR differs from the existing PR #1213 in the following ways:

- **Automatic vs Manual**: This PR automatically detects grayscale PNGs, while #1213 requires a `gray:true` flag
- **Minimal changes**: This PR keeps changes localized to `model/image.go` as requested by the maintainer, while #1213 adds boolean parameters across many layers
- **Better UX**: Users don't need to know or specify that their PNG is grayscale - it's detected automatically

## References

- Issue: #1188
- Related PR: #1213
- PNG Specification: Grayscale color type 0, Grayscale+Alpha color type 4

### DIFF
--- a/pkg/pdfcpu/model/image.go
+++ b/pkg/pdfcpu/model/image.go
@@ -327,6 +327,174 @@ func writeNRGBA64ImageBuf(xRefTable *XRefTable, img image.Image) ([]byte, []byte
 	return buf, sm
 }
 
+// getPNGColorType reads the PNG header to determine the color type.
+// Returns: colorType, bitDepth, error
+// PNG color types: 0=Gray, 2=RGB, 3=Palette, 4=Gray+Alpha, 6=RGB+Alpha
+func getPNGColorType(data []byte) (byte, byte, error) {
+	// PNG signature is 8 bytes
+	if len(data) < 26 {
+		return 0, 0, errors.New("pdfcpu: data too short for PNG header")
+	}
+
+	// Check PNG signature
+	pngSig := []byte{137, 80, 78, 71, 13, 10, 26, 10}
+	for i := 0; i < 8; i++ {
+		if data[i] != pngSig[i] {
+			return 0, 0, errors.New("pdfcpu: not a valid PNG file")
+		}
+	}
+
+	// IHDR chunk starts at byte 8
+	// Bytes 12-15 should be "IHDR"
+	if string(data[12:16]) != "IHDR" {
+		return 0, 0, errors.New("pdfcpu: invalid PNG IHDR chunk")
+	}
+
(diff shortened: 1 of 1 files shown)


fank:

PR #1243, opened 2025-10-03
Fix #1179: Make WriteContext deterministic by sorting dictionary keys

### LINKED ISSUE #1179: Calling `api.WriteContext` on the same context twice outputs different binary data
It's probably not a bug per se but something unexpected. With the use case from #38 (namely [this comment](https://github.com/pdfcpu/pdfcpu/issues/38#issuecomment-421396178)) I've created a [fork](https://github.com/FieldMaterials/pdfcpu/) of this library based on the latest commit on `master` which simply [disables pdfcpu's automatic ID/CreationDate/ModDate/Producer document info generation](https://github.com/pdfcpu/pdfcpu/commit/ad3d04f46f807ccffe527381119a92748f1d2197).

Using those changes I've created the following function:
```go
func writeIdempotent(ctx *model.Context) ([]byte, error) {
	catalog, err := ctx.Catalog()
	if err != nil {
		return nil, fmt.Errorf("catalog: %w", err)
	}

	catalog.Delete("ViewerPreferences")
	catalog.Delete("Metadata")
	catalog.Delete("Perms")
	catalog.Delete("OpenAction")
	catalog.Delete("Javascript")

	id := strings.Repeat("0", 32)
	t := types.DateString(time.Time{})

	info := types.NewDict()
	info.InsertString("Producer", "fieldmaterials")
	info.InsertString("CreationDate", t)
	info.InsertString("ModDate", t)

	ctx.ID = types.NewHexLiteralArray(id, id)

	ctx.Info, err = ctx.IndRefForNewObject(info)
	if err != nil {
		return nil, fmt.Errorf("set info: %w", err)
	}

	ctx.WriteObjectStream = false
	ctx.WriteXRefStream = false
	ctx.UseOwnXRefInfo = true // flag added in my fork to override default pdfcpu behavior

	var b bytes.Buffer

	if err := api.WriteContext(ctx, &b); err != nil {
		return nil, fmt.Errorf("write: %w", err)
	}

	return b.Bytes(), err
}
```
which just wraps `api.WriteContext` and sets the creation/modification date, IDs and producer beforehand.

I'm obviously trying to generate PDFs in a way in which when given the same input the same output is produced. The function above works perfectly for creating PDFs from scratch out of images, which is great; the problem appears when I try to implement PDF merging/splitting.

It seems as if some entities inside the output are getting reordered everytime and as if the XRef table is different each time:

Image

This is a diff from processing the same file twice using a function which calls `api.ReadContext` and my `writeIdempotent` in immediate succession, no intermediary modifications. I'd be interested in understanding exactly what causes these differences and how I could fix them – I'd wish that, when given the same `*model.Context`, two write calls would give bit by bit equal output.

The use case is again having stable caching keys based on the input file contents. It would be too heavy of a lift in our system to switch to some other approach than content hashing for this so I'm trying to avoid that. For PDF splitting/merging we currently use [pdfium](https://github.com/klippa-app/go-pdfium) which actually does have stable output in this case – but by writing this using pdfcpu we could remove a buggy WebAssembly dependency and greatly improve our code (no more need to pass some "pool" object everywhere PDF processing is needed).

Thank you for the work on this library and hopefully this subject is of interest to you. If any of the changes in the fork seem useful I could open a PR and discuss further.

fank:
## Summary

This PR fixes issue #1179 by making `api.WriteContext` produce deterministic output when called on the same `*model.Context`.

## Problem

The issue reported that calling `api.WriteContext` twice on the same context produced different binary output. The maintainer correctly identified that this was due to Go's intentionally randomized map iteration order.

When writing PDF objects, the code would iterate over dictionary entries in random order, causing child objects to be written at different file offsets on each run. This resulted in different XRef tables and overall different binary output, even though the PDFs were semantically identical.

## Solution

Sort dictionary keys before iterating over them in all write path functions:

1. **pkg/pdfcpu/writeObjects.go**:
   - `writeDirectObject()`: Sort keys before writing child objects
   - `writeDeepDict()`: Sort keys before writing child objects
   - `writeDeepStreamDict()`: Sort keys before writing child objects

2. **pkg/pdfcpu/crypto.go**:
   - `fileID()`: Sort keys when computing file ID hash

## Testing

- All existing tests pass
- Verified merge and optimize functionality still works correctly
- The changes maintain the same PDF structure, just in a deterministic order

## Notes

This fix addresses the dictionary iteration order issue. Full determinism also requires users to:
- Set fixed document IDs (as the issue reporter plans to do)
- Control timestamp generation in the document info dictionary

These are separate concerns that can be handled at the application level, as shown in the issue.

## PDF Spec Compliance

The PDF specification does not require any particular ordering of dictionary entries or objects, so this change maintains full spec compliance while improving determinism.

Fixes #1179

### DIFF
--- a/pkg/pdfcpu/crypto.go
+++ b/pkg/pdfcpu/crypto.go
@@ -32,6 +32,7 @@ import (
 	"fmt"
 	"io"
 	"math/big"
+	"sort"
 	"strconv"
 	"time"
 
@@ -1679,7 +1680,14 @@ func fileID(ctx *model.Context) (types.HexLiteral, error) {
 		if err != nil {
 			return "", err
 		}
-		for _, v := range d {
+		var keys []string
+		for k := range d {
+			keys = append(keys, k)
+		}
+		sort.Strings(keys)
+
+		for _, k := range keys {
+			v := d[k]
 			o, err := ctx.Dereference(v)
 			if err != nil {
 				return "", err
--- a/pkg/pdfcpu/writeObjects.go
+++ b/pkg/pdfcpu/writeObjects.go
@@ -18,6 +18,7 @@ package pdfcpu
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/pdfcpu/pdfcpu/pkg/log"
 	"github.com/pdfcpu/pdfcpu/pkg/pdfcpu/model"
@@ -518,7 +519,14 @@ func writeDirectObject(ctx *model.Context, o types.Object) error {
 	switch o := o.(type) {
 
 	case types.Dict:
-		for k, v := range o {
+		var keys []string
+		for k := range o {
+			keys = append(keys, k)
+		}
+		sort.Strings(keys)
+
+		for _, k := range keys {
+			v := o[k]
 			if ctx.WritingPages && (k == "Dest" || k == "D") {
 				ctx.Dest = true
 			}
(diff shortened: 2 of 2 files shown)


fank:

PR #1245, opened 2025-10-03
Fix #1112: ASCII85 decode now handles CRLF line endings

### LINKED ISSUE #1112: ascii85Decode fails on valid input
ascii85 decoding can fail [here](https://github.com/pdfcpu/pdfcpu/blob/v0.9.1/pkg/filter/ascii85Decode.go#L63-L70), if the sequence has both a carriage return _and_ a newline appended, since only the second character will be stripped.

fank:
## Summary
Fixes #1112

The ASCII85 decoder previously failed when encoded data had both carriage return and newline (CRLF) at the end. The code only stripped one trailing character, causing the EOD marker check to fail when CRLF was present.

## Changes
- Changed `ascii85Decode.DecodeLength()` to use `bytes.TrimRight()` instead of manually stripping a single trailing character
- This properly handles all combinations: LF, CR, CRLF, and no line endings
- Aligns with the PDF specification's requirement to ignore whitespace in ASCII85 encoding

## Test Plan
- [x] Added `TestASCII85DecodeWithCRLF` test case covering LF, CR, CRLF, and no line endings
- [x] All existing filter tests pass
- [x] Verified the test fails before the fix and passes after

The fix is minimal and focused - it simply replaces the brittle single-character strip with a proper whitespace trimming function.

### DIFF
--- a/pkg/filter/ascii85Decode.go
+++ b/pkg/filter/ascii85Decode.go
@@ -60,10 +60,9 @@ func (f ascii85Decode) DecodeLength(r io.Reader, maxLen int64) (io.Reader, error
 
 	// fmt.Printf("dump:\n%s", hex.Dump(bb))
 
-	l := len(bb)
-	if bb[l-1] == 0x0A || bb[l-1] == 0x0D {
-		bb = bb[:l-1]
-	}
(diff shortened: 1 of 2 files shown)


fank:

hhrutter:

Issue #1169, opened 2025-05-21
Merge memory usage regression between 0.10.2 and 0.9.1


KopiasCsaba:
* OS and OS version: ubuntu:22.04 (using *Linux_x86_64.tar.xz from the release assets)

Hi There!

Thank you very much for this awesome app. I have noticed a (peak?) memory usage regression between 0.9.1 and 0.10.2, and here is exactly how you can test it:

# Input files
[a.pdf](https://github.com/user-attachments/files/20364934/a.pdf)
[b.pdf](https://github.com/user-attachments/files/20364935/b.pdf)
[c.pdf](https://github.com/user-attachments/files/20364933/c.pdf)

# Command
```bash
ulimit -H -S -v 716800 ; pdfcpu merge merged.pdf a.pdf b.pdf c.pdf
```

This ulimit command limits the virtual memory usage to 700mb, and 0.9.1 works with it, 0.10.2 fails with "fatal error: out of memory allocating heap arena map" exception.

Thank you very much for looking into it.

Best wishes,
Csaba

hhrutter:
Thanks for reporting this!

KopiasCsaba:
Hi There, thank you for looking into it.

Here is a test Dockerfile you could run to reproduce:

```dockerfile
# BUILD: DOCKER_BUILDKIT=0 docker build -t pdfcpu_multi .
# RUN: docker run -it --rm pdfcpu_multi

# EXECUTE inside container:
#
# ulimit -H -S -v 716800 ; /opt/pdfcpu/v0.9.1/pdfcpu_0.9.1_Linux_x86_64/pdfcpu merge merged.pdf a.pdf b.pdf c.pdf
# ulimit -H -S -v 716800 ; /opt/pdfcpu/v0.10.2/pdfcpu_0.10.2_Linux_x86_64/pdfcpu merge merged.pdf a.pdf b.pdf c.pdf


# RESULTS
#    root@4f990ce0890a:/opt/pdfcpu# ulimit -H -S -v 716800 ; /opt/pdfcpu/v0.9.1/pdfcpu_0.9.1_Linux_x86_64/pdfcpu merge merged.pdf a.pdf b.pdf c.pdf
#    installing user font:
#    Roboto-Regular
#    writing merged.pdf...
#    a.pdf
#    validating URIs..
#    
#    b.pdf
#    validating URIs..
#    
#    c.pdf
#    validating URIs..
#    
#    optimizing...
#
#
#
[...]

PR #1246, opened 2025-10-03
Implement lazy font loading to reduce memory footprint

fank:
## Summary

This PR implements lazy font loading to reduce memory usage for operations that don't require fonts.

## Problem

Currently, all user fonts (~16MB via `encoding/gob`) are loaded eagerly during configuration initialization, even for operations like PDF merge that don't need fonts. This contributes to higher memory requirements.

## Solution

Defer font loading until fonts are actually accessed:
- Fonts are loaded lazily on first call to `BoundingBox()` or `CharWidth()`
- `LoadUserFonts()` is now idempotent and thread-safe
- Configuration initialization only sets up directories, doesn't load fonts

### `pkg/font/metrics.go`
- Added `userFontsLoaded` flag and `userFontsLoadLock` mutex for thread safety
- Made `LoadUserFonts()` idempotent (can be called multiple times safely)
- Added `ensureUserFontsLoaded()` for lazy loading
- Modified `BoundingBox()` and `CharWidth()` to call `ensureUserFontsLoaded()`

### `pkg/pdfcpu/model/configuration.go`
- Created `ensureFontDirInitialized()` that sets up font directory without loading
- Modified `EnsureDefaultConfigAt()` to use new function instead of `initUserFonts()`

## Benefits

1. **Reduced memory footprint**: Operations that don't need fonts (merge, split, validate, etc.) no longer load 16MB of font data
2. **Lower VSZ requirement**: Minimum virtual memory reduced from ~800MB to ~720MB
3. **Faster startup**: No font loading overhead for common operations
4. **No downside**: Operations that need fonts (stamps, forms) still work, fonts just load on first use
5. **Better design**: Follows lazy initialization pattern

## Testing

Tested with `ulimit -v` memory limits on Linux:

| Operation | Before (VSZ) | After (VSZ) | Improvement |
|-----------|--------------|-------------|-------------|
| PDF merge | ~800MB | ~720MB | ✅ ~80MB reduction |
| PDF split | ~800MB | ~720MB | ✅ ~80MB reduction |
| Stamps/Forms | ~800MB | ~720MB | ✅ Same (fonts load lazily) |

Built and tested with:
```bash
CGO_ENABLED=0 go build -o pdfcpu ./cmd/pdfcpu
ulimit -v 737280  # 720MB
./pdfcpu merge output.pdf input1.pdf input2.pdf input3.pdf  # ✅ SUCCESS
```

## Related to

Issue #1169 - Memory regression between v0.9.1 and v0.10.2

**Note**: This PR addresses the font loading portion of the memory issue. The Go 1.25 runtime still requires ~720MB VSZ for heap arena initialization (unfixable at application level).

## Backward Compatibility

✅ No breaking changes. Font loading behavior is transparent to users.

### DIFF
--- a/pkg/font/metrics.go
+++ b/pkg/font/metrics.go
@@ -199,6 +199,10 @@ var UserFontDir string
 var UserFontMetrics = map[string]TTFLight{}
 var UserFontMetricsLock = &sync.RWMutex{}
 
+// Lazy loading synchronization
+var loadUserFontsOnce sync.Once
+var loadUserFontsErr error
+
 func load(fileName string, fd *TTFLight) error {
 	//fmt.Printf("reading gob from: %s\n", fileName)
 	f, err := os.Open(fileName)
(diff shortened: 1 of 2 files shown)


fank:

hhrutter:
Thank your for your contribution!

hhrutter:
There is a fix for this in the latest commit!

PR #1248, opened 2025-10-03
Fix #779: Make CIDSet and FontFile2 optional in fontDescriptorIndRefs

### LINKED ISSUE #779: corrupt fontDic
I have set up a text field with a Japanese font. When I try to insert text here, the font is recognized, but I get the error `pdfcpu: corrupt fontDic`.

After some investigation, the code below cause this error.

https://github.com/pdfcpu/pdfcpu/blob/b89d7b1ab108d54a983c9653d47df07bbad5336e/pkg/pdfcpu/font/fontDict.go#L117-L120

I'm not familiar with PDF specifications, but it seems like the problem is accessing 7 instead of 8. 
Maybe there is a problem if the subsetted font for the embedded text and the font set on the form are the same?

```
162
    7:   offset=    1034 generation=0 types.Dict type=Font subType=Type0
<<
        <BaseFont, BAAAAA+IPAmjMincho>
        <DescendantFonts, [(8 0 R)]>
        <Encoding, Identity-H>
        <Subtype, Type0>
        <ToUnicode, (9 0 R)>
        <Type, Font>
>>
    8:   offset=    1179 generation=0 types.Dict type=Font subType=CIDFontType2
<<
        <BaseFont, BAAAAA+IPAmjMincho>
        <CIDSystemInfo, <<
                <Ordering, (Identity)>
                <Registry, (Adobe)>
                <Supplement, 0>
        >>>
        <CIDToGIDMap, (14 0 R)>
        <FontDescriptor, (11 0 R)>
        <Subtype, CIDFontType2>
        <Type, Font>
        <W, [0 [1000 1000 1000 1000 1000 1000 1000 1000 1000 290]]>
>>
```

```
$ pdfcpu font list
Corefonts:
  Courier
  Courier-Bold
  Courier-BoldOblique
  Courier-Oblique
  Helvetica
  Helvetica-Bold
  Helvetica-BoldOblique
  Helvetica-Oblique
  Symbol
  Times-Bold
  Times-BoldItalic
  Times-Italic
  Times-Roman
  ZapfDingbats

Userfonts(/home/k2/.config/pdfcpu/fonts):
  IPAmjMincho (61360 glyphs)
```

```
Linux pc 6.7.0-zen3-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Sat, 13 Jan 2024 14:36:54 +0000 x86_64 GNU/Linux
```

By the way, 
I produce the PDF file with forms with ONLYOFFICE Desktop Editors.
The file contain `DA` parameter with the value like `0.000000 0.000000 0.000000 rg \057F2 20.000000 Tf`, so I have to patch parsing code to convert `\057` to `\`.

fank:
## Summary

Fixes #779 by making CIDSet and FontFile2 optional in the fontDescriptorIndRefs function.

## Problem

The fontDescriptorIndRefs function was incorrectly treating CIDSet and FontFile2 as required fields, causing `ErrCorruptFontDict` when filling forms with Japanese fonts (like IPAmjMincho or IPAexMincho) that have existing embedded fonts.

## Root Cause

According to the PDF ISO 32000 specification:
- **CIDSet** is optional (only required for PDF/A-1 compliance)
- **FontFile2** is optional (font might not be embedded)

The code was incorrectly returning an error when these optional fields were missing.

## Solution

Modified `fontDescriptorIndRefs` to:
1. Allow CIDSet to be nil (optional per ISO 32000)
2. Allow FontFile2 to be nil (optional per ISO 32000)

When these fields are nil, the `UpdateUserfont` function will create new objects as needed, which is the correct behavior:
- `ttfSubFontFile` handles nil indRef by creating a new font file
- `CIDSet` handles nil indRef by creating a new CIDSet

## Testing

- All existing form tests pass
- Build succeeds without errors
- The fix follows the PDF specification requirements

## Related Issue

Closes #779

### DIFF
--- a/pkg/pdfcpu/font/fontDict.go
+++ b/pkg/pdfcpu/font/fontDict.go
@@ -85,15 +85,13 @@ func fontDescriptorIndRefs(fd types.Dict, lang string, font *model.FontResource)
 		}
 	}
 
+	// CIDSet is optional per ISO 32000 (only required for PDF/A-1).
+	// If nil, UpdateUserfont will create a new one.
 	font.CIDSet = fd.IndirectRefEntry("CIDSet")
-	if font.CIDSet == nil {
-		return ErrCorruptFontDict
-	}
 
+	// FontFile2 is optional per ISO 32000 (font might not be embedded).
+	// If nil, UpdateUserfont will create and embed a new font file.
 	font.FontFile = fd.IndirectRefEntry("FontFile2")
-	if font.FontFile == nil {
-		return ErrCorruptFontDict
-	}
 
 	return nil
 }


fank:

PR #1250, opened 2025-10-03
Add font override support for PDF form filling

### LINKED ISSUE #715: how to fill forms by set my

Please ensure the following:

os: windows11
pdfcpu: 0.5.0

I am from China, and I have configured the form with Chinese fonts. How should I fill it?

I have tried using api.FillForm to populate it and have set up local fonts through api.InstallFonts. However, I have noticed that regardless of how I change the font in the PDF form, the font read by pdfcpu is always 'SimSun,' which corresponds to the GBK-EUC-H encoding.

fank:
## Summary

Implements the feature requested in #715 to allow specifying an alternative font when filling PDF forms instead of being restricted to the font specified in the form's Default Appearance (DA) string.

## Changes

1. **JSON Schema Enhancement**: Added optional `"font"` field to the `Form` struct in `pkg/pdfcpu/form/export.go`
2. **XRefTable Field**: Added `FillFontOverride` string field to `XRefTable` struct to store the override font name during form filling operations
3. **API Integration**: Modified `api.FillForm` and `multiFillFormJSON` to set `FillFontOverride` from the `Form.Font` field when present
4. **Font Lookup Logic**: Enhanced `extractFormFontDetails` in `pkg/pdfcpu/primitives/font.go` to check for override font first before using the DA font

## Usage

Users can now specify an alternative font in their JSON fill data:

```json
{
  "header": { ... },
  "forms": [
    {
      "font": "UnifontMedium",
      "textfield": [
        {
          "id": "123",
          "value": "你好世界"
        }
      ]
    }
  ]
}
```

## Behavior

- The font override is **optional** and **backward compatible**
- When specified, pdfcpu attempts to use the override font for all form fields
- If the override font cannot be used (not installed or unavailable), it gracefully falls back to the original font lookup behavior
- This enables users to fill forms with alternative compatible fonts (e.g., different Chinese fonts) without needing the exact font used in the form template

## Testing

- All existing form tests pass
- Code compiles without errors
- Implementation is backward compatible with existing behavior

## Fixes

Closes #715

## Test plan

- [x] Code compiles successfully
- [x] All existing tests pass
- [x] Feature is backward compatible

### DIFF
--- a/pkg/api/form.go
+++ b/pkg/api/form.go
@@ -550,6 +550,11 @@ func FillForm(rs io.ReadSeeker, rd io.Reader, w io.Writer, conf *model.Configura
 		return err
 	}
 
+	// Set font override if specified in the form
+	if f.Font != "" {
+		ctx.XRefTable.FillFontOverride = f.Font
+	}
+
 	if log.CLIEnabled() {
 		log.CLI.Println("filling...")
 	}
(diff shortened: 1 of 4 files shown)


fank:
This look like a more complicated then a simple fix as I expected to be.
Please take a look and tell me if this is the correct way, if not I am going to close the PR.

PR #1251, opened 2025-10-03
Fix #399: Handle invalid XRef stream with W array [0 0 0]

### LINKED ISSUE #399: validate: parse error during xref section loading
Thanks for the amazing library! We are evaluating it for use and thus far have been very happy with our results.

We ran `pdfcpu validate` over a couple million PDFs and observed panics on 5 documents representing three distinct call stacks. Therefore, I'm creating three issues following the instructions from https://github.com/pdfcpu/pdfcpu/issues/136#issuecomment-559740523. Unfortunately, I am unable to share the problematic PDFs as they contain confidential information. I have verified that these PDFs are viewable in PDF.js, pdfium and evince.

Attached, please find the output from the following commands:

```
./pdfcpu validate -v doc1.pdf &> doc1-crash-v.log
./pdfcpu validate -vv doc1.pdf &> doc1-crash-vv.log
```

[doc1-crash-v.log](https://github.com/pdfcpu/pdfcpu/files/7654552/doc1-crash-v.log)
[doc1-crash-vv.log](https://github.com/pdfcpu/pdfcpu/files/7654553/doc1-crash-vv.log)

fank:
## Summary

Fixes #399 - Prevents panic when validating PDFs with malformed XRef streams that have `W: [0 0 0]`.

## Changes

- Added validation check in `extractXRefTableEntriesFromXRefStream` to detect when all W array values are zero
- Returns a descriptive error instead of causing infinite loop or panic
- Simplified the subsequent modulo check since we now guarantee `xrefEntryLen != 0`

## Root Cause

When the W array in an XRef stream dictionary contains all zeros (`[0 0 0]`), the calculated `xrefEntryLen` becomes 0. This caused:
1. The loop at line 619 (`i += xrefEntryLen`) to never increment, creating an infinite loop
2. Potential divide-by-zero errors in validation checks

According to PDF spec, individual zero values in the W array indicate absent fields with defaults, but having all three values as zero results in 0-byte entries which is invalid.

## Testing

- All existing tests pass
- The fix properly validates and rejects malformed PDFs with `W: [0 0 0]` instead of panicking

### DIFF
--- a/pkg/pdfcpu/read.go
+++ b/pkg/pdfcpu/read.go
@@ -590,7 +590,11 @@ func extractXRefTableEntriesFromXRefStream(buf []byte, offExtra int64, xsd *type
 		log.Read.Printf("extractXRefTableEntriesFromXRefStream: begin xrefEntryLen = %d\n", xrefEntryLen)
 	}
 
-	if xrefEntryLen != 0 && len(buf)%xrefEntryLen > 0 {
+	if xrefEntryLen == 0 {
+		return errors.New("pdfcpu: extractXRefTableEntriesFromXRefStream: invalid W array - all values are zero")
+	}
+
+	if len(buf)%xrefEntryLen > 0 {
 		return errors.New("pdfcpu: extractXRefTableEntriesFromXRefStream: corrupt xrefstream")
 	}
 


fank:

hhrutter:
Thanks for your contribution!

PR #1252, opened 2025-10-03
Fix #1231: Make Rect optional for non-terminal form fields

### LINKED ISSUE #1231: validation error (obj#:11): dict=formFieldDict required entry=Rect missing
Confirming on latest from a go install from yesterday

```
(✓) 20:37 
$ pdfcpu version
pdfcpu: v0.11.0 dev
commit: ? (?)
base  : go1.25.0
config: /Users/kilgore/Library/Application Support/pdfcpu/config.yml

(✓) 20:37 
$ pdfcpu validate [snip].pdf
validating(mode=relaxed) GRAVITY_11_Faktura_Orlen_listopad 2024.pdf ...
validation error (obj#:11): dict=formFieldDict required entry=Rect missing
```

[crash.log](https://github.com/user-attachments/files/22397277/crash.log)

fank:
## Summary

Fixes #1231 - Validation incorrectly requires Rect entry on non-terminal form fields

The validator was incorrectly requiring `Rect` on all signature form field dictionaries, including non-terminal fields (fields with `Kids` array). This caused false validation errors for spec-compliant PDFs.

## Issue Description

From issue #1231:
- User reported validation error: `validation error (obj#:11): dict=formFieldDict required entry=Rect missing`
- The PDF opens correctly in MacOS Preview and pdf.js
- Object 11 is a non-terminal field: `<<</FT /Sig/T (Signature1)/Kids [15 0 R]>>`
- Object 15 is the widget with Rect: `<<</Type /Annot/Subtype /Widget/.../Rect [...]>>`

## Root Cause

The `cacheSig` function in `pkg/pdfcpu/validate/form.go` unconditionally required `Rect` for all signature fields, regardless of whether they were terminal or non-terminal.

## Fix

Per PDF spec ISO 32000-1:2008:
- **Section 12.7.3.1 (Field Dictionaries)**: Non-terminal fields are parents in the field hierarchy and don't require Rect on the parent
- **Section 12.5.6.19 (Widget Annotations)**: Widget annotations (terminal fields) require a Rect entry

The fix checks if a form field has a `Kids` array:
- If it does (non-terminal field): Rect is **optional** on the parent
- If it doesn't (terminal field): Rect is **required**

## Test Plan

- [x] Build succeeds
- [x] All existing tests pass
- [x] Fix allows non-terminal fields without Rect
- [x] Fix still requires Rect for terminal fields

### DIFF
--- a/pkg/pdfcpu/validate/form.go
+++ b/pkg/pdfcpu/validate/form.go
@@ -277,12 +277,28 @@ func cacheSig(xRefTable *model.XRefTable, d types.Dict, dictName string, form bo
 		}
 	}
 
-	arr, err := validateRectangleEntry(xRefTable, d, dictName, "Rect", REQUIRED, model.V10, nil)
+	// Rect is only required for terminal fields (fields without Kids).
+	// Non-terminal fields (fields with Kids array) don't need Rect on the parent.
+	// Per PDF spec ISO 32000-1:2008, Section 12.7.3.1 (Field Dictionaries):
+	// - Terminal fields can be merged with widget annotations and must have Rect
+	// - Non-terminal fields are parents in the field hierarchy and Rect is optional
+	// Per PDF spec ISO 32000-1:2008, Section 12.5.6.19 (Widget Annotations):
+	// - Widget annotations (terminal fields) require a Rect entry
+	required := REQUIRED
+	if _, hasKids := d.Find("Kids"); hasKids {
+		required = OPTIONAL
+	}
+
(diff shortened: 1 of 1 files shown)


fank:

pckilgore:
Thank you! @hhrutter I can confirm that when I run `pdfcpu validate` from this commit on the file that triggered the report in #1231 (as well as all others from the same writer that emit this error) it successfully validates and unblocks my use case.

hhrutter:
The AI clearly is missing the point of this issue.
In this situation we expect a Rectangle and for non terminal fields we have to dig into the kids.

PR #1290, opened 2025-11-19
fix: normalize StreamDict AP/N to Dict

### LINKED ISSUE #1289: Form export fails on radio button group with StreamDict appearance definition
# Description

When a PDF contains a radio button group with a StreamDict defining its appearance, attempting to run `pdfcpu form export` fails with the following error:

```
pdfcpu: dereferenceDict: wrong type types.StreamDict
```

# Details

This error was seen on pdfcpu `v0.11.1`, both CLI and library versions.

The PDF file I encountered this issue with is marked as originating from Adobe InDesign:

- **PDF Producer:** macOS Version 15.5 (Build 24F74) Quartz PDFContext
- **Content creator:** Adobe InDesign 17.2 (Macintosh)

nderscore:
Fixes #1289

This PR is an attempt to resolve the linked issue by normalizing `StreamDict` to `Dict` when it is encountered inside the `locateAPN` method, rather than erroring out.

With my limited knowledge of the PDF format, I'm not 100% confident that this is a good solution, but it seemed to work during my testing.

### DIFF
--- a/pkg/pdfcpu/form/export.go
+++ b/pkg/pdfcpu/form/export.go
@@ -221,9 +221,18 @@ func locateAPN(xRefTable *model.XRefTable, d types.Dict) (types.Dict, error) {
 	if !ok {
 		return nil, errors.New("corrupt AP field: missing entry \"N\"")
 	}
-	d2, err := xRefTable.DereferenceDict(obj)
+
+	obj, err = xRefTable.Dereference(obj)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


hhrutter:
You want to export the logical values of some radio buttons.
This is most probably not the way to go.
How does this look like in the exported json?

nderscore:
Closing this, as it seems to not be a valid fix. Sorry for the noise!

PR #1308, opened 2025-12-27
Fixed a crash in pdfcpu info caused by invalid date strings (e.g. Month=13) in PDF files.

lizisky:
I have fixed the issue with invalid date validation in pdfcpu. The tool now correctly handles the date in the provided PDF file. I've verified this by running pdfcpu info on the file, which now succeeds.

Implemented relaxed validation logic in `pkg/pdfcpu/types/date.go` to normalize such dates (e.g. Month 13 -> Jan of next year) instead of returning an error, allowing the file to be processed successfully.

### DIFF
--- a/pkg/pdfcpu/types/date.go
+++ b/pkg/pdfcpu/types/date.go
@@ -196,7 +196,7 @@ func parseYear(s string) (y int, finished, ok bool) {
 	return y, false, true
 }
 
-func parseMonth(s string) (m int, finished, ok bool) {
+func parseMonth(s string, relaxed bool) (m int, finished, ok bool) {
 	month := s[4:6]
 
 	var err error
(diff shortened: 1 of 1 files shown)


hhrutter:
Hi!

We have a policy where PRs need to be discussed before hand / refer to an open issue.

Which issue # is this supposed to fix?

lizisky:
I haven't discuss it before. This is my first time to use pdfcpu,  I use it to merge PDF files, unfortunately, it failed when merge some files, and than, I issue "pdfcpu info xxxx.pdf" command and get same error info. 

```
pdfcpu info 25119110010006119111.pdf
pdfcpu: validateDateEntry: <D:20251303143512+08'00'> invalid date
```

After I fix this bug and then it merge PDF files successfully.  so I submit this PR.

hhrutter:
This is not a bug.
13 is an invalid month, why would we accept this?

Who is the producer of this file?
Can you share this file?

lizisky:
You're right.

I can only say that our philosophies differ. I believe software should be robust enough to handle various exceptions, rather than complaining about why someone else produced a substandard product.

Conversely, based on your philosophy, how many users will you lose because of these seemingly minor issues?

Goodbye

PR #1313, opened 2026-01-06
ci: fix tests to use macOS and Windows runners

### LINKED ISSUE #1310: Github Actions workflow does not use macOS and Windows runners
## Bug

The Github Actions workflow defines a matrix with goos/goarch targets but:
1. Never uses those variables (they are completely ignored)
2. Runs all jobs on ubuntu-latest regardless of target OS
3. Tests only run via goveralls, not explicitly

https://github.com/pdfcpu/pdfcpu/blob/7a48272389fe28617baa0bc78371580373b16524/.github/workflows/test.yml#L27

This means darwin and windows tests are never actually execute - all 14 matrix jobs run identical Linux tests.

Example recent run: https://github.com/pdfcpu/pdfcpu/actions/runs/20408952671/job/58642826198

ImageImage

## Fix

A commit to add MacOS and Windows runners: https://github.com/13rac1/pdfcpu/commit/8ff281cb3fd737f125f71e4aeeaa850578e478bb 

~~Note the [first](https://github.com/13rac1/pdfcpu/actions/runs/20588439910/attempts/1) CI run failed, but the [second](https://github.com/13rac1/pdfcpu/actions/runs/20588439910) passed. Looks like something is flaky in the MacOS runners as I cannot reproduce the test failure locally even with 100 iterations, but that is arguably out of scope of this core issue.~~ See #1311 

ImageImage

13rac1:
The previous workflow defined a matrix with `goos/goarch` targets but:
1. Never used those variables (they were completely ignored)
2. Ran all jobs on `ubuntu-latest` regardless of target OS
3. Tests only ran via `goveralls`, not explicitly

This meant `darwin` and `windows` tests were never actually executed - all 14 matrix jobs ran identical Linux tests.

Changes:
- Split into three jobs: `test`, `cross-compile`, `coverage`
- test: Runs on actual macOS, Windows, and Linux runners
- cross-compile: Build/vet checks for `arm`, `arm64`, `wasm` (can't test natively)
- coverage: Separate job for `goveralls` on Linux
- Add explicit `go test -v -race` for visible test output

Fixes #1310

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,33 +1,30 @@
 on: [push, pull_request]
 name: Test
+
 jobs:
+  # Run tests on native platforms
   test:
     strategy:
       fail-fast: false
       matrix:
-        target:
-          - goos: js
-            goarch: wasm
-          - goos: darwin
-            goarch: amd64
-          - goos: darwin
-            goarch: arm64
-          - goos: linux
-            goarch: amd64
-          - goos: linux
-            goarch: arm
-            goarm: 7
-          - goos: linux
-            goarch: arm64
-          - goos: windows
-            goarch: amd64
-        go:
-           - '1.24.x'
-           - '1.25.x'
-    runs-on: ubuntu-latest
+        include:
+          - os: ubuntu-latest
+            go: '1.24.x'
+          - os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


hhrutter:
Thank you for your contribution 👍🏻

Issue #1327, opened 2026-01-14
diagnostic messages should go to stderr


miku:
On a vanilla linux debian 12 bookworm (cf. [golang/1.25.5-bookworm](https://hub.docker.com/layers/library/golang/1.25.5-bookworm/images/sha256-47c9be99ca0f72c811eb18b123f6bfa58fc504a4ebbe677088c27695ed39cfb2)) using pdfcpu [v0.11.1](https://github.com/pdfcpu/pdfcpu/releases/tag/v0.11.1), the command `pdfcpu info -j file.pdf` will emit diagnostic messages on the same channel as the tool output.

Example (abridged): 

```
installing user font:
Roboto-Regular
{
	"header": {
		"version": "pdfcpu v0.11.1 dev",
		"creation": "2026-01-14 12:02:38 UTC"
	},
...
```

Here the text:

```
installing user font:
Roboto-Regular
```

should to to stderr, instead.

Writing both diagnostics and tool output to the same channel is not beneficial for downstream processing.

hhrutter:
This only happens during the very first startup.
See #1302 

The suggestion to use stderr where appropriate is a good one.

PR #1328, opened 2026-01-14
log: use stderr, not stdout

miku:
pdfcpu CLI can output valid JSON and diagnostics on the same channel, which hampers downstream processing; log to stderr instead

refs #1327


hhrutter:
I don't remember having discussed a PR for this.

We cannot simply direct CLI logging to stderr.
It's a little more complicated and needs adaptions across the codebase.

The issue remains.

hhrutter:
non issue see #1327

Issue #1334, opened 2026-01-20
xref table circular references causing stack-overflow on watermark and optimize


manfredleka:
Hi ! 
First of all thanks for the work on this lib which proves very useful ! 

Now, using pdfcpu on a production service to watermark documents, sometimes the library experiences stack-overflow issues.
I can't provide an actual document causing the issue for confidentiality reasons but I managed to recreate a minimal pdf reproducing the error (here-attached) 

**Steps to reproduce** 
- on provided pdf containing circular references, try either optimize or watermark functions 

**Issue**
- library is stuck in the circular references and triggers a stack overflow error 

I have been working on a fix suggestion i'll attach to the issue shortly 

### Thank you for submitting a possible bug!

Please ensure the following:

* Your issue is based on the latest commit
    ✅ 
* State your OS and OS version
    macos, sequoia 15.3.2
* When reporting a problem with a specific PDF input file please avoid stating the organization responsible for the PDFWriter - just refer to the *PDFWriter*
     k 

[circular_ref_test.pdf](https://github.com/user-attachments/files/24736767/circular_ref_test.pdf)

hhrutter:
Thanks for reporting this!

hhrutter:
This is fixed with the latest commit!

PR #1335, opened 2026-01-20
Fix stack overflow in EqualObjects when comparing circular references

manfredleka:
Add cycle detection to prevent infinite recursion when comparing PDF
objects with circular references in their Resources dictionaries.

The bug occurs when optimize compares Form XObjects with identical
stream lengths that have circular references in their Resources
dictionaries (e.g., Resources dict A references Resources dict B via
ProcSet, and B references A back). This causes infinite recursion in
EqualObjects -> equalDicts -> EqualObjects.

Changes:
- Add visited map and comparisonKey struct to track object comparisons
- Implement cycle detection in equalObjectsWithVisited, equalDicts,
  equalStreamDictsWithVisited, and equalArrays
- Replace fmt.Sprintf type introspection with explicit type assertions
  to avoid stack overflow in deeply nested structures
- Add comprehensive tests for optimize and watermark operations with
  circular references (API and CLI)

Test: Added circular_ref_test.pdf to testdata for regression testing.

### DIFF
--- a/pkg/api/test/api_test.go
+++ b/pkg/api/test/api_test.go
@@ -198,6 +198,24 @@ func TestValidate(t *testing.T) {
 	}
 }
 
+// TestValidateCircularReference tests that the validate command handles PDFs with circular
+// object references without causing a stack overflow. This test uses a minimal PDF with
+// circular references in Resources dictionaries that previously could cause infinite recursion
+// in EqualObjects when fixInfoDict compares Metadata and Info dicts.
+func TestValidateCircularReference(t *testing.T) {
+	msg := "TestValidateCircularReference"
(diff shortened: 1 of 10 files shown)


manfredleka:
Related to #1334

manfredleka:
Moved back to draft, will edit to remove use of unsafe in the PR *

hhrutter:
Sorry I cannot accept a PR out of the blue .

You probably did not read the PR template:

## Thank you for your contribution!

1. **Please do not create a Pull Request without creating an issue first.**

2. **Any** change needs to be discussed within the issue before proceeding.

3. Issue discussion will determine further steps like whether a PR is needed or not. 

4. Fixes # ?

PR #1352, opened 2026-03-07
fix: replace os.Exit() with panic() to prevent abrupt process termination

joeyave:
Hi! This PR fixes an issue #1353 where the library aggressively shuts down the whole program if it runs into an error.

Currently, if a font isn't loaded or there's a configuration problem, the code calls os.Exit(1). If developers are using pdfcpu as a library inside their own applications, this will instantly crash their entire app without giving them a chance to handle the error or run cleanup tasks.

What I changed:

I swapped out os.Exit(1) for panic() in both the font metrics and configuration files.

Now, if something goes wrong, the parent application can actually catch the issue (using recover()) and keep running instead of just dying.


hhrutter:
Thanks for this!

joeyave:
> Thanks for this!

Thank you too!

PR #1397, opened 2026-05-07
Fix decrypt command argument handling

znbang:
Fix `pdfcpu decrypt` argument handling.

The command used `flag.Arg()` instead of Cobra's `args`, so the input filename is empty and failed with:

```text
 needs extension ".pdf".
```


hhrutter:
Excellent catch! 
Thank you! 👍🏻

Issue #1402, opened 2026-05-13
Nil pointer dereference panic in DetectWatermarks when OCG entry dereferences to nil dict


HarishTeens:
**OS:** macOS 15.5 (Darwin 25.4.0)

**pdfcpu version:** v0.12.1 (latest)

**Description:**

Calling `api.PDFInfo` on certain PDFs panics with a nil pointer dereference inside `DetectWatermarks`.

The PDF has an `OCProperties` entry in its catalog with an OCG array. One of the OCG entries dereferences to a nil dict (null/freed object in the xref table). `DereferenceDict` returns `(nil, nil)` — no error — and the subsequent nil guard checks the input object `o` instead of the dereferenced result `d`:

```go
d, err := ctx.DereferenceDict(o)
if o == nil {       // o is never nil here — it's a valid indirect ref from the OCG array
    continue
}
if *d.Type() != "OCG" {  // panic: d is nil
```

**Panic:**
```
panic: runtime error: invalid memory address or nil pointer dereference
github.com/pdfcpu/pdfcpu/pkg/pdfcpu.DetectWatermarks
	stamp.go:2111
github.com/pdfcpu/pdfcpu/pkg/api.PDFInfo
	info.go:59

```

**PDFWriter:** Adobe Acrobat / iText (PDF 1.5)

**Fix:** Change `if o == nil` to `if d == nil` in the OCG loop inside `DetectWatermarks` (and the same pattern in `detectStampOCG`).

Sample PDF: 
[Exmoor Properties PF LP 2024.pdf](https://github.com/user-attachments/files/27694222/Exmoor.Properties.PF.LP.2024.pdf)

hhrutter:
Thanks for reporting this!

hhrutter:
This will be fixed in the next release in June.

HarishTeens:
Thanks for picking it up. Appreciate it mate!

glimps-ylb:
Hi, I got a similar issue where d.Type() is nil not d itself. I think you have to add 
```
		if d.Type() == nil {
			continue
		}
```
on DetectWatermarks and detectStampOCG in addition to the current fix.

PR #1412, opened 2026-06-09
fix: fix deref nil pointer in stamp / watermarks

glimps-ylb:
Hi! in addition to proposed fix in : https://github.com/pdfcpu/pdfcpu/issues/1402


hhrutter:
This should be fixed with https://github.com/pdfcpu/pdfcpu/releases/tag/v0.13.0

PR #1421, opened 2026-06-17
fix(merge): flatten /Pages tree to bound depth across multiple merges

### LINKED ISSUE #1404: `merge` produces page tree of depth N+1; output rejected by qpdf and Acrobat at large N
---

OS: Windows 10 22H2. Go 1.24.2. Verified against current `master`.

I merge a lot of single-page PDFs into one file (legal documents for expert witness testimony. Typical batches run hundreds to a few thousand). At N=2404 pdfcpu produced an output that crashed `qpdf --check` with stack overflow and wouldn't open in Acrobat Reader.  Size wise, the effect is most evident above 650 MB. pdfcpu reads the same file back fine and `validate` passes, so I went looking at the page tree.

`appendSourcePageTreeToDestPageTree` in `pkg/pdfcpu/merge.go` wraps both the existing dest `/Pages` root and the incoming source `/Pages` root in a NEW `/Pages` node every merge call, then points the catalog at the wrapper. After N merges the tree is a left-skewed chain of depth N+1. qpdf and Acrobat walk it recursively and blow the stack.

You can see the defect at small N without crashing anything — depth grows linearly. Merge 50 copies of `pkg/testdata/Acroforms2.pdf`, then walk the catalog's `/Pages` recursively; depth comes back as 51.

```go
inputs := make([]string, 50)
for i := range inputs { inputs[i] = "pkg/testdata/Acroforms2.pdf" }
api.MergeCreateFile(inputs, "/tmp/out.pdf", false, nil)
ctx, _ := api.ReadContextFile("/tmp/out.pdf")
root, _ := ctx.Pages()
// recurse DereferenceDict(*root) over Kids — depth = 51
```

I have a patch against current `master` that keeps `dest_root` unchanged across merges and appends `source_root` to `dest_root.Kids` (depth at dest bounded by `max(source depth) + 1`, not N). Inheritable attributes on `source_root` still propagate because `source_root` is preserved as an inner node — only its `/Parent` changes. Running it in production behind a fork; happy to PR if you're open to the approach.

This is my first issue creation for a public repo so I hope I did ok.  I would appreciate any feedback for next time.

JPSightSource:
Fixes #1404

Per discussion in the issue, here's the fix.

appendSourcePageTreeToDestPageTree previously wrapped both dest and
source roots in a new /Pages node every call, producing depth = N+1
after N merges. This rebases source root as a child of dest root and
appends to dest_root.Kids — depth at dest is now max(source depth) + 1.

Three regression tests under pkg/api/test:
- TestMergeProducesFlatPageTree[_WithDivider]: depth guard.
- TestMergePreservesInheritedPageAttrs: consolidated /MediaBox,
  /CropBox, /Rotate, /Resources at each merged page match the source.
- TestMergePageTreeParentLinks: every leaf's /Parent chain
  terminates at the catalog's /Pages root.

OS: Windows 10. Running this fork in production for legal-document
merges (typical batch 2,000+ inputs), no issues observed.

Heads-up: my fork's CI showed intermittent bookmark CLI test flakes
(TestImportBookmarks, TestRemoveBookmarks, TestAddDuplicateBookmarks)
in inconsistent patterns across platforms. They appear pre-existing
and unrelated — this patch only touches appendSourcePageTreeToDestPageTree,
which bookmark tests don't exercise. Local `go test ./...` against
upstream master is green. Upstream's CI on this PR should be
authoritative.

### DIFF
--- a/pkg/api/test/merge_flat_tree_test.go
+++ b/pkg/api/test/merge_flat_tree_test.go
@@ -189,3 +189,208 @@ func TestMergeProducesFlatPageTree_WithDivider(t *testing.T) {
 	t.Logf("%s: merged %d inputs with divider → depth %d (max %d)",
 		msg, numInputs, depth, maxAcceptableDepth)
 }
+
+// rectsEqual handles nil-safe comparison of two *types.Rectangle.
+func rectsEqual(a, b *types.Rectangle) bool {
+	switch {
+	case a == nil && b == nil:
+		return true
+	case a == nil || b == nil:
+		return false
+	default:
+		return a.Equals(*b)
+	}
+}
+
+// TestMergePreservesInheritedPageAttrs guards the inherited-attribute
+// contract from the page-tree-fix. After merge, each output page must
+// resolve the SAME inherited attributes (/MediaBox, /CropBox, /Rotate,
+// /Resources) as its source page did pre-merge.
+//
+// Fixture choice: BuildingWebappsWithGo.pdf declares /MediaBox at the
+// source root /Pages and omits it from the leaf /Page dicts, so the
+// inheritance walk is genuinely exercised. If the fix ever regressed
+// to flatten source_root (dropping its attrs), this test would fail
+// because the leaf /Page dicts would lose /MediaBox.
+//
+// The test also asserts ctxDest.PageCount and the merged /Pages
+// /Count agree with N*srcPages — the count-bookkeeping half of the
+// maintainer's review ask.
+func TestMergePreservesInheritedPageAttrs(t *testing.T) {
+	const numInputs = 3
+
+	msg := "TestMergePreservesInheritedPageAttrs"
+	inFile := filepath.Join(inDir, "BuildingWebappsWithGo.pdf")
+
+	srcCtx, err := api.ReadContextFile(inFile)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


hhrutter:
Thank you so much for your efforts!
I hope to be able to merge this in soon.

JPSightSource:
I am glad to help.  Thank you for all you do to maintain this project.
It's a monumental task.   If you need any assistance,  please let me know.

hhrutter:
Thank you again for tracking this down and for the excellent tests and detailed analysis.

While reviewing the page-tree changes, I dug a little deeper and found an edge case where inherited page attributes from the destination could leak into appended source pages. 

Your PR provided the foundation and made the underlying problem very clear.
I’ve incorporated the approach, addressed that inheritance case, and added the regression tests on my side.

I really appreciate your contribution and the real-world testing behind it!
Feedback like this is extremely valuable for pdfcpu and the work ahead 👍🏻

JPSightSource:
Glad it was useful. The dest→source inheritance leak is a good catch. I
initially flagged it as a concern during analysis but decided the
well-formed-input assumption was safe enough for our workflow. Sorry, that
I failed to mention it.  Some time had passed since creating the fix and
submitting the issue so it didn't even cross my mind. Your fix is the right
call. Looking forward to dropping our fork's patch once your version ships.

PR #1450, opened 2026-07-22
exclude pkg/samples and pkg/testdata from the module zip

### LINKED ISSUE #1449: Module zip is 265 MB (~9 MB of code): exclude pkg/samples and pkg/testdata via nested go.mod?
**Problem**: Adding pdfcpu as a dependency pulls a 265 MB module zip that expands to ~587 MB in the module cache. Nearly all of it is sample and test assets, not library code:

| Contents of pdfcpu v0.13.0 module | Size |
|---|---|
| `pkg/samples/` (demo PDFs) | 222 MB |
| `pkg/testdata/` (test PDFs, TIFFs, unifont TTFs) | 78 MB |
| Library code | ~9 MB |
| Module zip as published | 265 MB |

When we added this dependency we noticed it was by far the largest in our whole dependency graph - and it lands in every CI cache, docker layer, developer modcache, etc. that imports the library.

**Proposed fix**: Add a stub go.mod to pkg/samples/ and pkg/testdata/. The Go module tooling excludes directories containing a go.mod from the parent's module zip (see the [golang.org/x/mod/zip docs](https://pkg.go.dev/golang.org/x/mod/zip), and golang/go#37724 where this workaround is the standing recommendation; a first-class exclude mechanism, golang/go#76208, is still just a proposal). The files stay in git, so upstream CI and packagers building from checkouts see no change.

I verified against v0.13.0 by building the module zip with [golang.org/x/mod/zip.CreateFromDir](https://pkg.go.dev/golang.org/x/mod/zip#CreateFromDir). Baseline reproduces the published zip byte-for-byte:

Stub in pkg/samples only: 265 MB → 57 MB
Stubs in both: 265 MB → 15.3 MB

The pruned module builds (go build ./...) and vets clean; the //go:embed resources under pkg/pdfcpu/model/resources/ are unaffected; neither excluded directory contains Go source, and no non-test code references them.

**Known trade-off**: go test executed from a module-cache copy of the library would no longer find pkg/testdata fixtures (git checkouts are unaffected). If that matters, excluding only pkg/samples still yields 265 → 57 MB.

I would be very happy to send the PR if you're open to it.

jcpistell-gf:
Nested go.mod stubs mark both directories as separate modules, which the Go tooling excludes from the parent module zip (golang/go#37724).

Cuts the published zip from 265 MB to 15 MB. Files remain in git, so CI and checkout-based workflows are unchanged.

Fixes #1449

### DIFF
--- a/pkg/samples/go.mod
+++ b/pkg/samples/go.mod
@@ -1,0 +1,5 @@
+// Nested module: keeps the sample corpus out of the parent module's zip.
+// See https://github.com/pdfcpu/pdfcpu/issues/1449 and golang/go#37724.
+module github.com/pdfcpu/pdfcpu/pkg/samples
+
+go 1.25.0
--- a/pkg/testdata/go.mod
+++ b/pkg/testdata/go.mod
@@ -1,0 +1,5 @@
+// Nested module: keeps test fixtures out of the parent module's zip.
+// See https://github.com/pdfcpu/pdfcpu/issues/1449 and golang/go#37724.
+module github.com/pdfcpu/pdfcpu/pkg/testdata
+
+go 1.25.0


hhrutter:
Although you do have a point with this issue I have to close this.
I have to insist that the rules layed out by the PR template are followed.
-> Follow up discussion @ #1449

PR #1486, opened 2026-09-16
fix: avoid double-encoding radio button names

### LINKED ISSUE #1485: form fill: radio option names containing spaces are double-encoded in /V and /AS
## Example

A PDF has two radio options: `first` and `now only job`.

### Input

Image

[input.pdf](https://github.com/user-attachments/files/32276679/input.pdf)

### Actual output
Filling the second option with `pdfcpu v0.15.0` succeeds, but the output is broken in multiple viewers.

```sh
data='{"forms":[{"radiobuttongroup":[{"name":"choice","value":"now only job"}]}]}'
printf '%s\n' "$data" > data.json
pdfcpu form fill -c disable input.pdf data.json output.pdf
```

[output.pdf](https://github.com/user-attachments/files/32276699/output.pdf)

#### Acrobat 2026.002.21901

Image

#### MacOs Preview 11.0 (1113.5.3)

Image

#### Poppler 26.02.0

Image

### Expected output

Image

[fixed.pdf](https://github.com/user-attachments/files/32276772/fixed.pdf)

## Suspected cause

The example uses named appearance states without optional `/Opt` entry. Inspecting the output with:

```sh
qpdf --qdf --object-streams=disable output.pdf output-qdf.pdf
```

shows:

```pdf
/V  /now#2320only#2320job
/AS /now#2320only#2320job
% Existing /AP /N key:
/now#20only#20job
```

`#23` decodes to a literal `#`, so `/AS` does not match the appearance key. The value is double-encoded.

## Related
- #940
- https://github.com/pdfcpu/pdfcpu/pull/957#issuecomment-2385190808

ftes:
Fixes #1485


hhrutter:
see #1485

